PR #17, opened 2019-07-12
Optimize UInt256 read/write on PostgreSQL

thebevrishot:


src/Ztm.Data.Entity.Postgres/Ztm.Data.Entity.Postgres.csproj
```diff
@@ -1,5 +1,10 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
+  <PropertyGroup>
+    <Version Condition=" '$(Version)' == '' ">4.1.2.35</Version>
+    <LangVersion>7.2</LangVersion>
```
ultimaweapon: Put this line in below `PropertyGroup` will be more cleaner.

src/Ztm.Data.Entity.Testing/TestMainDatabase.cs
```diff
@@ -19,6 +19,52 @@ protected override void ConfigureWebApiCallback(ModelBuilder modelBuilder)
...
+
+            modelBuilder.Entity<WatchingTransaction>(b =>
+            {
+                b.Property(e => e.Hash).IsRequired().HasConversion(Converters.UInt256ToBytesConverter);
+            });
+
+            modelBuilder.Entity<WebApiCallback>(b =>
+            {
+                b.Property(e => e.TransactionId).HasConversion(Converters.UInt256ToBytesConverter);
+            });
```
ultimaweapon: If we want to put all of these logic in a single method we should override `OnModelCreating` instead of `ConfigureWebApiCallback`.

src/Ztm.Data.Entity.Postgres/TypeMappingSource.cs
```diff
@@ -0,0 +1,37 @@
+using Microsoft.EntityFrameworkCore.Storage;
+using NBitcoin;
+using Npgsql.EntityFrameworkCore.PostgreSQL.Infrastructure.Internal;
+using Npgsql.EntityFrameworkCore.PostgreSQL.Storage.Internal;
+using Ztm.Data.Entity.Postgres.Mapping;
+
+namespace Ztm.Data.Entity.Postgres
+{
+    public class TypeMappingSource : NpgsqlTypeMappingSource
```
ultimaweapon: This class should be internal due to it not designed to use outside this assembly.

src/Ztm.Data.Entity.Postgres/Mapping/UInt256TypeMapping.cs
```diff
@@ -0,0 +1,33 @@
+using System.Linq.Expressions;
+using Microsoft.EntityFrameworkCore.Storage;
+using NBitcoin;
+using Npgsql.EntityFrameworkCore.PostgreSQL.Storage.Internal.Mapping;
+using NpgsqlTypes;
+
+namespace Ztm.Data.Entity.Postgres.Mapping
+{
+    public class UInt256TypeMapping : NpgsqlTypeMapping
```
ultimaweapon: Same here.

src/Ztm.Data.Entity.Postgres/MainDatabase.cs
```diff
@@ -1,10 +1,25 @@
...
 {
     public class MainDatabase : Ztm.Data.Entity.Contexts.MainDatabase
     {
+        static MainDatabase()
+        {
+            NpgsqlConnection.GlobalTypeMapper.AddMapping(new NpgsqlTypeMappingBuilder
+            {
+                PgTypeName = "bytea",
+                NpgsqlDbType = NpgsqlDbType.Bytea,
+                ClrTypes = new[] { typeof(uint256), typeof(byte[]) },
```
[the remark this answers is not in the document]
thebevrishot: nope, we need this, without this parsing bytea to []byte is broken.

src/Ztm.Data.Entity.Postgres/ByteaHandler.cs
```diff
@@ -1,66 +1,34 @@
...
-        #region Read
-        public override byte[] Read(NpgsqlReadBuffer buf, int len, FieldDescription fieldDescription = null)
+        async ValueTask<uint256> INpgsqlTypeHandler<uint256>.Read(NpgsqlReadBuffer buf, int len, bool async, FieldDescription fieldDescription)
         {
             try
             {
                 var bytes = new byte[len];
-                buf.ReadBytes(bytes, 0, len);
-                return bytes;
+                await buf.ReadBytes(bytes, 0, len, async);
```
ultimaweapon: We should use `Npgsql.TypeHandlers.ByteaHandler` to handle `bytea` instead by calling `base.Read`. Also should throw if data length is not exactly 32 bytes.

src/Ztm.Data.Entity.Postgres/ByteaHandler.cs
```diff
@@ -1,66 +1,34 @@
...
+                return new uint256(bytes, false);
```
ultimaweapon: Currently we store `uint256` in the database as little endian, so this one is not correct that passing second parameter as false.

src/Ztm.Data.Entity.Postgres/ByteaHandler.cs
```diff
@@ -1,66 +1,34 @@
...
             }
             catch (Exception e)
             {
                 throw new NpgsqlSafeReadException(e);
             }
         }
 
-        protected override uint256 ReadPsv(NpgsqlReadBuffer buf, int len, FieldDescription fieldDescription = null)
+        public async Task Write(uint256 value, NpgsqlWriteBuffer buf, NpgsqlLengthCache lengthCache, NpgsqlParameter parameter, bool async)
```
ultimaweapon: We should explicit implement this method so it is consistency with `Read` that we just implemented above.

src/Ztm.Data.Entity.Postgres/ByteaHandler.cs
```diff
@@ -1,66 +1,34 @@
...
-            {
-                var bytes = new byte[len];
-                buf.ReadBytes(bytes, 0, len);
-                return new uint256(bytes, storeUInt256AsLE);
-            }
-            catch (Exception e)
-            {
-                throw new NpgsqlSafeReadException(e);
-            }
+            await buf.WriteBytesRaw(value.ToBytes(false), async);
```
ultimaweapon: Same as previous issue. We should let `bytea` handle by built-in handler by calling `base.Write` instead.

ultimaweapon: `value.ToBytes(false)` is not correct since currently we store it as little-endian.

src/Ztm.Data.Entity.Postgres/ByteaHandler.cs
```diff
@@ -9,26 +9,19 @@ namespace Ztm.Data.Entity.Postgres
...
-            {
-                var bytes = new byte[len];
-                await buf.ReadBytes(bytes, 0, len, async);
-                return new uint256(bytes, false);
-            }
-            catch (Exception e)
-            {
-                throw new NpgsqlSafeReadException(e);
-            }
+            var bytes = await Read<byte[]>(buf, len, async, fieldDescription);
```
ultimaweapon: This one is quite dangerous due to this overload was designed to be able to read arbitrary type. We should use https://github.com/npgsql/npgsql/blob/dev/src/Npgsql/TypeHandlers/ByteaHandler.cs#L38 instead.

PR #28, opened 2019-10-08
Fix lgtm warning

thebevrishot:


src/Ztm.Data.Entity/Contexts/Main/Input.cs
```diff
@@ -41,5 +41,35 @@ public int CompareTo(Input other)
...
+
+            return CompareTo((Input)other) == 0;
+        }
+
+        public override int GetHashCode()
+        {
+            int hash = 0;
+
+            hash ^= (TransactionHash != null) ? TransactionHash.GetHashCode() : 0;
+            hash ^= (int)Index;
```
ultimaweapon: After discussion we decided to turn off this LGTM alert and use `Index.GetHashCode()` instead.

src/Ztm.Data.Entity/Contexts/Main/Input.cs
```diff
@@ -41,5 +41,35 @@ public int CompareTo(Input other)
...
+            hash ^= (OutputHash != null) ? OutputHash.GetHashCode() : 0;
```
ultimaweapon: We don't need this lines and below due to this class represent an input on a transaction. So just `TransactionHash` and `Index` should be enough.

src/Ztm.Data.Entity/Contexts/Main/Output.cs
```diff
@@ -41,5 +41,33 @@ public int CompareTo(Output other)
 
             return 0;
         }
+
+        public override bool Equals(Object other)
```
ultimaweapon: Same as previous issue.

src/Ztm.Data.Entity/Contexts/Main/Output.cs
```diff
@@ -41,5 +41,33 @@ public int CompareTo(Output other)
...
+        {
+            if (other == null)
+            {
+                return false;
+            }
+
+            if (GetType() != other.GetType())
```
ultimaweapon: Same as previous issue.

src/Ztm.Data.Entity/Contexts/Main/Output.cs
```diff
@@ -41,5 +41,33 @@ public int CompareTo(Output other)
...
+
+            return CompareTo((Output)other) == 0;
+        }
+
+        public override int GetHashCode()
+        {
+            int hash = 0;
+
+            hash ^= (TransactionHash != null) ? TransactionHash.GetHashCode() : 0;
+            hash ^= (int)Index;
```
ultimaweapon: Same as previous issue.

src/Ztm.Data.Entity/Contexts/Main/Output.cs
```diff
@@ -41,5 +41,33 @@ public int CompareTo(Output other)
...
+            hash ^= (int)Value;
```
ultimaweapon: Same as previous issue for this line and below.

src/Ztm.Zcoin.Synchronization/AddressWatcher.cs
```diff
@@ -158,9 +158,8 @@ public class AddressWatcher : ITransactionConfirmationListener
                 }
 
                 // Add debit to address.
-                balances[address] = balances.ContainsKey(address)
-                    ? balances[address] + spend.Value
-                    : spend.Value;
+                Money val;
+                balances[address] = balances.TryGetValue(address, out val) ? val + spend.Value : spend.Value;
```
ultimaweapon: We can do something like this `balances.TryGetValue(address, out var val)` and remove the above line.

ultimaweapon: IMO better to rename `val` to `previous` will be better.

src/Ztm.Data.Entity/Contexts/Main/Input.cs
thebevrishot: Make sense

src/Ztm.Zcoin.Synchronization/AddressWatcher.cs
thebevrishot: Good to know this sugar syntax

thebevrishot: make sense

src/Ztm.Data.Entity/Contexts/Main/Block.cs
```diff
@@ -32,27 +32,22 @@ public int CompareTo(Block other)
...
+            }
 
-                return Hash == other.Hash ? 0 : -1;
+            if (Hash != null)
+            {
+                return Hash.CompareTo(other.Hash);
             }
 
-            return Height - other.Height;
+            return Hash == other.Hash ? 0 : -1;
```
ultimaweapon: We need to modify this every time we add more fields to compare. My idea is change the above `if` to:

```cs
if (Hash != null)
{
    return Hash.CompareTo(other.Hash);
}
else if (other.Hash != null)
{
    return -1;
}
```

Then we can change this line to just `return 0;`.

Issue #34, opened 2019-10-09
Move all of additional public members of ZcoinBlockHeader to be extensions method


ultimaweapon:
This will remove casting `BlockHeader` to `ZcoinBlockHeader` every time we need to access additional members.

PR #36, opened 2019-10-09
Add header extension

thebevrishot:
resolve https://github.com/zcoinofficial/ztm/issues/34


src/Ztm.Zcoin.NBitcoin/ZcoinBlock.cs
```diff
@@ -19,9 +19,9 @@ public ZcoinBlock(ZcoinConsensusFactory consensusFactory, ZcoinBlockHeader heade
         }
         #pragma warning restore CS0618
 
-        public new ZcoinBlockHeader Header
+        public new BlockHeader Header
```
ultimaweapon: We don't need this anymore.

Issue #47, opened 2019-10-15
Implement callback status


ultimaweapon:
Contains the following properties:

- `Status` as string.
- `Data` as object.

It need to have constant for `success`, `update` and `error` status too.

- [ ] Implementation [0.5]
- [ ] Review [0.5]

PR #50, opened 2019-10-16
Add callback result

thebevrishot:
resolves https://github.com/zcoinofficial/ztm/issues/47

### DIFF
--- a/src/Ztm.WebApi/CallbackResult.cs
+++ b/src/Ztm.WebApi/CallbackResult.cs
@@ -2,20 +2,11 @@ namespace Ztm.WebApi
 {
     public abstract class CallbackResult
     {
-        public static class CallbackStatus
-        {
-            public const string Error   = "error";
-            public const string Success = "success";
-            public const string Update  = "update";
-        }
+        public const string StatusError   = "error";
+        public const string StatusSuccess = "success";
+        public const string StatusUpdate  = "update";
 
-        public CallbackResult(string status, object data)
-        {
-            this.Status = status;
-            this.Data = data;
-        }
-
-        public string Status { get; }
-        public object Data { get; }
+        public abstract string Status { get; }
+        public abstract object Data { get; }
     }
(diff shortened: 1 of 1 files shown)


src/Ztm.WebApi/CallbackResult.cs
```diff
@@ -0,0 +1,21 @@
...
+            public const string Update  = "update";
+        }
+
+        public CallbackResult(string status, object data)
+        {
+            this.Status = status;
+            this.Data = data;
+        }
+
+        public string Status { get; }
```
ultimaweapon: If we choose to make it as abstract class we should make this abstract.

src/Ztm.WebApi/CallbackResult.cs
```diff
@@ -0,0 +1,21 @@
...
+        public object Data { get; }
```
ultimaweapon: Same here.

src/Ztm.WebApi/CallbackResult.cs
```diff
@@ -0,0 +1,21 @@
...
+    public abstract class CallbackResult
+    {
+        public static class CallbackStatus
+        {
+            public const string Error   = "error";
+            public const string Success = "success";
+            public const string Update  = "update";
+        }
+
+        public CallbackResult(string status, object data)
```
ultimaweapon: We don't need this constructor.

src/Ztm.WebApi/CallbackResult.cs
```diff
@@ -0,0 +1,21 @@
+namespace Ztm.WebApi
+{
+    public abstract class CallbackResult
+    {
+        public static class CallbackStatus
+        {
+            public const string Error   = "error";
+            public const string Success = "success";
+            public const string Update  = "update";
```
ultimaweapon: We should move all above to the parent class and suffix it with `Status`.

PR #73, opened 2019-10-24
Decoupled watchers from IBlockListener

### LINKED ISSUE #61: Refactor ConfirmationWatcher to make it as a concrete class

ultimaweapon:
After investigation we cannot make it as a concrete class due to it need to determine the conditions to create watch.

This PR will remove watcher framework to act as background service and change it storage to handler instead. Example use case for watcher framework is:

- Create a class to do something when some watches are completed (e.g. do something when a specific transaction got 6 confirmations).
- That class need to implement a specific watcher handler for the watchers it want to use (e.g. transaction confirmation watcher).
- That class need implement `IBlockListener` in order to receive notifications when block was added or removed. Than invoke `ExecuteAsync` on each watcher it want.

Resolve #61.


ultimaweapon:
Don't review this yet. Got another idea to work on.

ultimaweapon:
Ready for review.

Issue #79, opened 2019-10-28
Implement Exodus transaction


ultimaweapon:
- Have a base class for all Exodus transaction data.
- Each transaction type will have a dedicated class derived from a base class.
- Class for each transaction type will contains only decoded payload.
- Add a new property to `ZcoinTransaction` for Exodus transaction data with a type of transaction data base class.
- A property will have a `null` value if that transaction is not Exodus transaction.
- [x] Implementation [3]
- [ ] Review [1]

PR #93, opened 2019-10-28
Move all Exodus related from Ztm.Zcoin.NBitcoin to Ztm.Zcoin.NBitcoin.Exodus

ultimaweapon:
First part of #79.


ultimaweapon:
We can ignore the new alerts. It already suppressed but don't know why it still comment about it.

PR #96, opened 2019-11-08
Implement base classes for Exodus transaction

### LINKED ISSUE #79: Implement Exodus transaction

ultimaweapon:
This not include a new property in `ZcoinTransaction` due to the PR will too large. We can add it later.

Resolve #79.

### DIFF
--- a/src/Ztm.Zcoin.NBitcoin.Tests/Exodus/TransactionDecoderTests.cs
+++ b/src/Ztm.Zcoin.NBitcoin.Tests/Exodus/TransactionDecoderTests.cs
@@ -1,0 +1,236 @@
+using System;
+using System.IO;
+using System.Net;
+using System.Text;
+using NBitcoin;
+using Xunit;
+using Ztm.Zcoin.NBitcoin.Exodus;
+
+namespace Ztm.Zcoin.NBitcoin.Tests.Exodus
+{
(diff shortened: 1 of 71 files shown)


src/Ztm.Zcoin.NBitcoin.Tests/Exodus/TransactionDecoderTests.cs
```diff
@@ -0,0 +1,236 @@
...
+        [InlineData(9)]
+        [InlineData(10)]
+        [InlineData(11)]
+        [InlineData(12)]
+        [InlineData(13)]
+        [InlineData(14)]
+        [InlineData(15)]
+        public void Decode_SimpleSendV0WithTooShortData_ShouldThrow(int length)
+        {
+            var payload = new byte[length - 4];
```
thebevrishot: Please add some description why 4 bytes are skipped.

Issue #94, opened 2019-11-04
Integrate API callback with ASP.NET


ultimaweapon:
The scope of this issue is:

- Extract callback information from the request.
- Write callback information into response.

See specification for information how callback works.

- [x] Implementation [2]
- [ ] Review [1]

PR #102, opened 2019-11-12
Add controller base extensions to get set callback data

thebevrishot:
resolves https://github.com/zcoinofficial/ztm/issues/94


src/Ztm.WebApi/HttpResponseExceptionFilter.cs
```diff
@@ -0,0 +1,36 @@
...
+    {
+        public void OnActionExecuting(ActionExecutingContext context)
+        {
+        }
+
+        public void OnActionExecuted(ActionExecutedContext context)
+        {
+            if (context.Exception is HttpResponseException exception)
+            {
+                context.Result = new JsonResult(new ErrorDetails{Message = exception.Value})
```
ultimaweapon: We should not use `JsonResult` due to it will break ASP.NET automatic response type.

ultimaweapon: The error response should compliance with [RFC 7807](https://tools.ietf.org/html/rfc7807). See here for example: https://docs.microsoft.com/en-us/aspnet/core/web-api/handle-errors?view=aspnetcore-2.2#exception-handler

src/Ztm.WebApi/HttpResponseExceptionFilter.cs
```diff
@@ -0,0 +1,36 @@
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.Filters;
+using Newtonsoft.Json;
+
+namespace Ztm.WebApi
+{
+    class ErrorDetails
```
ultimaweapon: If we follow my idea seem like we don't need this.

src/Ztm.WebApi/InvalidCallbackUrlException.cs
```diff
@@ -0,0 +1,24 @@
+using System.Net;
+
+namespace Ztm.WebApi
+{
+    public sealed class InvalidCallbackUrlException : HttpResponseException
```
ultimaweapon: We need to update this once we update `HttpResponseException`.

src/Ztm.WebApi/ControllerBaseExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+namespace Ztm.WebApi
+{
+    public static class ControllerBaseExtensions
+    {
+        readonly static string CallbackUrlKey = "X-Callback-URL";
+        readonly static string CallbackIdKey = "X-Callback-ID";
+
+        public static bool TryGetCallbackUrl(this ControllerBase controller, out Uri url)
+        {
+            if (controller.HttpContext.Request.Headers.TryGetValue(CallbackUrlKey, out var rawUrl))
```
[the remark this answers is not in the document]
thebevrishot: Nice catch

src/Ztm.WebApi/ControllerBaseExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+                {
+                    url = new Uri(rawUrl);
+
+                    if (url.Scheme == Uri.UriSchemeHttp
+                        || url.Scheme == Uri.UriSchemeHttps)
+                    {
+                        return true;
+                    }
+
+                    throw new InvalidCallbackUrlException(rawUrl);
```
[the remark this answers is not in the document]
thebevrishot: nice catch again!!

src/Ztm.WebApi/ControllerBaseExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+                {
+                    throw new InvalidCallbackUrlException(rawUrl);
+                }
+            }
+
+            url = null;
+            return false;
+        }
+
+        public static void SetCallbackId(this ControllerBase controller, Guid id)
```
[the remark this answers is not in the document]
thebevrishot: Already try to create an overload but doesn't work then I use `AcceptedWithCallback` instead.

src/Ztm.WebApi/HttpResponseExceptionFilter.cs
thebevrishot: Create controller to handle error instead

src/Ztm.WebApi/HttpResponseExceptionFilter.cs
thebevrishot: removed

src/Ztm.WebApi/InvalidCallbackUrlException.cs
thebevrishot: done

src/Ztm.WebApi/Controllers/ErrorController.cs
```diff
@@ -0,0 +1,47 @@
...
+                Instance = feature?.Path,
+                Title = isHttpResponseException ? $"{((HttpResponseException)ex).Title} : {ex.GetType().Name}" : ex.GetType().Name,
+                Detail = $"{ex.Message} : {ex.StackTrace}",
+            };
+
+            return StatusCode(problemDetails.Status.Value, problemDetails);
+        }
+
+        [Route("/error")]
+        public ActionResult Error([FromServices] IHostingEnvironment webHostEnvironment)
```
ultimaweapon: `webHostEnvironment` never referenced.

src/Ztm.WebApi/Controllers/ErrorController.cs
```diff
@@ -0,0 +1,47 @@
...
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Mvc;
+
+namespace Ztm.WebApi.Controllers
+{
+    [ApiController]
+    public class ErrorController : ControllerBase
+    {
+        [Route("/error-development")]
+        public ActionResult ErrorDevelopment([FromServices] IHostingEnvironment webHostEnvironment)
```
ultimaweapon: Same here.

src/Ztm.WebApi/Controllers/ErrorController.cs
```diff
@@ -0,0 +1,47 @@
...
+                Detail = $"{ex.Message} : {ex.StackTrace}",
+            };
+
+            return StatusCode(problemDetails.Status.Value, problemDetails);
+        }
+
+        [Route("/error")]
+        public ActionResult Error([FromServices] IHostingEnvironment webHostEnvironment)
+        {
+            var feature = HttpContext.Features.Get<IExceptionHandlerPathFeature>();
```
ultimaweapon: `feature` is quite confusing. IMO rename to `info` looks better.

src/Ztm.WebApi/Controllers/ErrorController.cs
```diff
@@ -0,0 +1,47 @@
...
+            var ex = feature?.Error;
+            var isHttpResponseException = ex is HttpResponseException;
+
+            var problemDetails = new ProblemDetails
```
ultimaweapon: I think the initialization of this looks ugly. My idea is just initialize the shared property first. Then assign `Status` and `Title` from exception if it is `HttpResponseException`. If not, assign it with `HttpStatusCode.InternalServerError` and `"An error occurred."`.

src/Ztm.WebApi/Controllers/ErrorController.cs
```diff
@@ -0,0 +1,47 @@
...
+
+namespace Ztm.WebApi.Controllers
+{
+    [ApiController]
+    public class ErrorController : ControllerBase
+    {
+        [Route("/error-development")]
+        public ActionResult ErrorDevelopment([FromServices] IHostingEnvironment webHostEnvironment)
+        {
+            var feature = HttpContext.Features.Get<IExceptionHandlerPathFeature>();
```
ultimaweapon: Same as previous issue.

src/Ztm.WebApi/Controllers/ErrorController.cs
```diff
@@ -0,0 +1,47 @@
...
+            var ex = feature?.Error;
+            var isHttpResponseException = ex is HttpResponseException;
+
+            var problemDetails = new ProblemDetails
```
ultimaweapon: Same as previous issue.

src/Ztm.WebApi/InvalidCallbackUrlException.cs
```diff
@@ -1,24 +1,20 @@
+using System;
 using System.Net;
 
 namespace Ztm.WebApi
 {
     public sealed class InvalidCallbackUrlException : HttpResponseException
     {
-        public InvalidCallbackUrlException(string InvalidUrl)
+        static readonly string InvalidCallbackUrlTitle = "Invalid Callback Url";
```
ultimaweapon: Rename to just `StaticTitle` looks better due to we already in `InvalidCallbackUrlException` class.

ultimaweapon: We should not apply naming convention on display text. My idea is change it value to `Invalid callback URL.`.

src/Ztm.WebApi/InvalidCallbackUrlException.cs
```diff
@@ -1,24 +1,20 @@
...
+        public InvalidCallbackUrlException()
+            : base((int)HttpStatusCode.BadRequest, InvalidCallbackUrlTitle)
         {
-            this.InvalidUrl = InvalidUrl;
         }
 
-        public override int Status => (int)HttpStatusCode.BadRequest;
-
-        public override object Value
+        public InvalidCallbackUrlException(Exception ex)
```
ultimaweapon: Same as previous issue.

ultimaweapon: This overload should have `string message` too so it match with the parent class.

src/Ztm.WebApi/Controllers/ErrorController.cs
```diff
@@ -9,35 +9,65 @@ namespace Ztm.WebApi.Controllers
...
+        public ActionResult Error()
         {
-            var feature = HttpContext.Features.Get<IExceptionHandlerPathFeature>();
-            var ex = feature?.Error;
-            var isHttpResponseException = ex is HttpResponseException;
+            var info = HttpContext.Features.Get<IExceptionHandlerPathFeature>();
+            var ex = info?.Error;
+
+            int status;
+            string title;
```
ultimaweapon: We can remove the above 2 variables by moving instantiation of `ProblemDetails` here. e.g.:

```csharp
var details = new ProblemDetails()
{
    Instance = info?.Path,
    Detail = null
};
```

src/Ztm.WebApi/Controllers/ErrorController.cs
```diff
@@ -9,35 +9,65 @@ namespace Ztm.WebApi.Controllers
...
+
+            if (ex is ApiException)
+            {
+                var apiEx = (ApiException)ex;
```
ultimaweapon: We can remove this cast by update `if` to `if (ex is ApiException apiEx)`.

src/Ztm.WebApi/Controllers/ErrorController.cs
```diff
@@ -9,35 +9,65 @@ namespace Ztm.WebApi.Controllers
...
+        public ActionResult ErrorDevelopment()
         {
-            var feature = HttpContext.Features.Get<IExceptionHandlerPathFeature>();
-            var ex = feature?.Error;
-            var isHttpResponseException = ex is HttpResponseException;
+            var info = HttpContext.Features.Get<IExceptionHandlerPathFeature>();
+            var ex = info?.Error;
+
+            int status;
+            string title;
```
ultimaweapon: Same as previous issue.

src/Ztm.WebApi/Controllers/ErrorController.cs
```diff
@@ -9,35 +9,65 @@ namespace Ztm.WebApi.Controllers
...
+
+            if (ex is ApiException)
+            {
+                var apiEx = (ApiException)ex;
```
ultimaweapon: Same as previous issue.

src/Ztm.WebApi/ControllerBaseExtensions.cs
```diff
@@ -19,7 +19,7 @@ public static bool TryGetCallbackUrl(this ControllerBase controller, out Uri url
                 }
                 catch (UriFormatException ex)
                 {
-                    throw new InvalidCallbackUrlException(ex);
+                    throw new InvalidCallbackUrlException(null, ex);
```
ultimaweapon: We should provide exception message so we know what happening. Even though it not used to response to client but it will be helpful in our log.

src/Ztm.WebApi/Controllers/ErrorController.cs
thebevrishot: I think current way is easier to read.

Issue #89, opened 2019-10-28
Implement RPC to get Exodus balance


ultimaweapon:
- [x] Implementation [0.5]
- [ ] Review [0.5]

Issue #83, opened 2019-10-28
Implement RPC for simple send


ultimaweapon:
- [x] Implementation [0.5]
- [x] Review [0.5]

PR #110, opened 2019-11-14
Add rpc to send token and get balance

thebevrishot:
closes https://github.com/zcoinofficial/ztm/issues/89 and https://github.com/zcoinofficial/ztm/issues/83

### DIFF
--- a/src/Ztm.Zcoin.Rpc.Tests/ZcoinRpcClientTests.cs
+++ b/src/Ztm.Zcoin.Rpc.Tests/ZcoinRpcClientTests.cs
@@ -459,7 +459,7 @@ namespace Ztm.Zcoin.Rpc.Tests
         }
 
         [Fact]
-        public async Task SendAsync_WithNullAsRequiredArgs_ShouldThrow()
+        public async Task SendTokenAsync_WithNullAsRequiredArgs_ShouldThrow()
         {
             var from = TestAddress.Regtest1;
             var to = TestAddress.Regtest2;
@@ -469,22 +469,22 @@ namespace Ztm.Zcoin.Rpc.Tests
 
             await Assert.ThrowsAsync<ArgumentNullException>(
                 "from",
-                () => this.subject.SendAsync(null, to, property, amount, null, null, CancellationToken.None)
+                () => this.subject.SendTokenAsync(null, to, property, amount, null, null, CancellationToken.None)
             );
 
             await Assert.ThrowsAsync<ArgumentNullException>(
                 "to",
-                () => this.subject.SendAsync(from, null, property, amount, null, null, CancellationToken.None)
+                () => this.subject.SendTokenAsync(from, null, property, amount, null, null, CancellationToken.None)
             );
 
             await Assert.ThrowsAsync<ArgumentNullException>(
                 "property",
-                () => this.subject.SendAsync(from, to, null, amount, null, null, CancellationToken.None)
+                () => this.subject.SendTokenAsync(from, to, null, amount, null, null, CancellationToken.None)
             );
         }
 
         [Fact]
-        public async Task SendAsync_WithInvalidArgs_ShouldThrow()
+        public async Task SendTokenAsync_WithInvalidArgs_ShouldThrow()
         {
             var from = TestAddress.Regtest1;
             var to = TestAddress.Regtest2;
@@ -494,12 +494,12 @@ namespace Ztm.Zcoin.Rpc.Tests
 
             await Assert.ThrowsAsync<ArgumentOutOfRangeException>(
                 "amount",
-                () => this.subject.SendAsync(from, to, property, PropertyAmount.Zero, null, null, CancellationToken.None)
+                () => this.subject.SendTokenAsync(from, to, property, PropertyAmount.Zero, null, null, CancellationToken.None)
             );
 
             await Assert.ThrowsAsync<ArgumentOutOfRangeException>(
                 "referenceAmount",
-                () => this.subject.SendAsync(from, to, property, amount, redeemAddress, Money.Satoshis(-1), CancellationToken.None)
+                () => this.subject.SendTokenAsync(from, to, property, amount, redeemAddress, Money.Zero, CancellationToken.None)
             );
 
         }
@@ -516,7 +516,7 @@ namespace Ztm.Zcoin.Rpc.Tests
             var destination = await this.subject.GetNewAddressAsync(CancellationToken.None);
 
             // Act.
-            var rawTx = await this.subject.SendAsync(property.Owner, destination, property.Property, new PropertyAmount(10), null, null, CancellationToken.None);
+            var rawTx = await this.subject.SendTokenAsync(property.Owner, destination, property.Property, new PropertyAmount(10), null, null, CancellationToken.None);
             await this.subject.SendRawTransactionAsync(rawTx, CancellationToken.None);
             this.node.Generate(1);
 
@@ -524,8 +524,8 @@ namespace Ztm.Zcoin.Rpc.Tests
             var ownerBalance = await this.subject.GetPropertyBalanceAsync(property.Owner, property.Property, CancellationToken.None);
             var destinationBalance = await this.subject.GetPropertyBalanceAsync(destination, property.Property, CancellationToken.None);
 
-            Assert.Equal(new PropertyAmount(990), ownerBalance.Balance);
-            Assert.Equal(new PropertyAmount(10), destinationBalance.Balance);
+            Assert.Equal(new PropertyAmount(990), ownerBalance.balance);
+            Assert.Equal(new PropertyAmount(10), destinationBalance.balance);
         }
 
         [Fact]
@@ -564,8 +564,8 @@ namespace Ztm.Zcoin.Rpc.Tests
                 property.Owner, property.Property, CancellationToken.None);
 
             // Assert.
-            Assert.Equal(new PropertyAmount(1000), balance.Balance);
-            Assert.Equal(new PropertyAmount(0), balance.Reserved);
+            Assert.Equal(new PropertyAmount(1000), balance.balance);
+            Assert.Equal(new PropertyAmount(0), balance.reserved);
         }
 
         [Fact]
@@ -577,15 +577,15 @@ namespace Ztm.Zcoin.Rpc.Tests
             this.node.Generate(101);
             var property = await issuer.IssueManagedAsync();
 
-            await property.GrantAsync(new PropertyAmount(10_0000_0000));
+            await property.GrantAsync(PropertyAmount.FromDivisible(10));
 
             // Act.
             var balance = await this.subject.GetPropertyBalanceAsync(
                 property.Owner, property.Property, CancellationToken.None);
 
             // Assert.
-            Assert.Equal(PropertyAmount.FromDivisible(10), balance.Balance);
-            Assert.Equal(PropertyAmount.FromDivisible(0), balance.Reserved);
+            Assert.Equal(PropertyAmount.FromDivisible(10), balance.balance);
+            Assert.Equal(PropertyAmount.FromDivisible(0), balance.reserved);
         }
 
         class ManagedProperty
@@ -599,13 +599,13 @@ namespace Ztm.Zcoin.Rpc.Tests
                 this.client = client;
             }
 
-            public async Task GrantAsync(PropertyAmount propertyAmount)
+            public async Task GrantAsync(PropertyAmount amount)
             {
                 var grantTx = await this.client.GrantPropertyAsync(
                     this.Property,
                     this.Owner,
                     this.Owner,
-                    propertyAmount,
+                    amount,
                     null,
                     CancellationToken.None);
 
@@ -620,16 +620,6 @@ namespace Ztm.Zcoin.Rpc.Tests
 
         class PropertyIssuer
         {
-            public BitcoinAddress Owner { get; set; }
-            public Ecosystem Ecosystem { get; set; }
-            public PropertyType PropertyType { get; set; }
-            public Property PropertyCurrent { get; set; }
-            public string Catagory { get; set; }
-            public string SubCatagory { get; set; }
-            public string Name { get; set; }
-            public string Url { get; set; }
-            public string Description { get; set; }
-
             readonly CoreNode node;
             readonly ZcoinRpcClient client;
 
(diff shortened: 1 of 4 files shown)


src/Ztm.Zcoin.Rpc.Tests/ZcoinRpcClientTests.cs
```diff
@@ -456,5 +456,226 @@ public async Task SendToAddressAsync_WithNullAmount_ShouldThrow()
...
+
+            Assert.Equal(new PropertyAmount(990), ownerBalance.Balance);
+            Assert.Equal(new PropertyAmount(10), destinationBalance.Balance);
+        }
+
+        [Fact]
+        public async Task GetPropertyBalanceAsync_WithNullArgs_ShouldThrow()
+        {
+            // Arrange.
+            var builder = new PropertyBuilder(this.node, this.subject);
```
ultimaweapon: Same as previous issue.

src/Ztm.Zcoin.Rpc.Tests/ZcoinRpcClientTests.cs
```diff
@@ -456,5 +456,226 @@ public async Task SendToAddressAsync_WithNullAmount_ShouldThrow()
...
+
+            // Assert.
+            await Assert.ThrowsAsync<ArgumentNullException>(
+                "address",
+                () => this.subject.GetPropertyBalanceAsync(null, managedProperty.property, CancellationToken.None)
+            );
+
+            await Assert.ThrowsAsync<ArgumentNullException>(
+                "property",
+                () => this.subject.GetPropertyBalanceAsync(address, null, CancellationToken.None)
```
ultimaweapon: Why we don't use token owner instead?

src/Ztm.Zcoin.Rpc.Tests/ZcoinRpcClientTests.cs
```diff
@@ -456,5 +456,226 @@ public async Task SendToAddressAsync_WithNullAmount_ShouldThrow()
...
+            );
+        }
+
+        [Fact]
+        public async Task GetPropertyBalanceAsync_WithValidArgs_ShouldSuccess()
```
ultimaweapon: Better to rename this case so it match with next case.

src/Ztm.Zcoin.Rpc.Tests/ZcoinRpcClientTests.cs
```diff
@@ -456,5 +456,226 @@ public async Task SendToAddressAsync_WithNullAmount_ShouldThrow()
...
+        {
+            // Arrange.
+            var builder = new PropertyBuilder(this.node, this.subject);
```
ultimaweapon: Same as previous issue.

src/Ztm.Zcoin.Rpc.Tests/ZcoinRpcClientTests.cs
```diff
@@ -456,5 +456,226 @@ public async Task SendToAddressAsync_WithNullAmount_ShouldThrow()
...
+            public BitcoinAddress owner { get; set; }
+            public Property property { get; set; }
+
+            public ManagedProperty(CoreNode node, ZcoinRpcClient client)
+            {
+                this.node = node;
+                this.client = client;
+            }
+
+            public async Task GrantAsync(long grantingAmount)
```
ultimaweapon: Why we not accept `PropertyAmount` instead of `long`?

src/Ztm.Zcoin.Rpc/ZcoinRpcClient.cs
```diff
@@ -241,6 +277,62 @@ public Task<uint256> SendRawTransactionAsync(Transaction tx, CancellationToken c
...
+                ToNative(property.Id),
+                amount.ToString(property.Type),
+            };
+
+            if (redeemAddress != null) {
+                args.Add(redeemAddress.ToString());
+
+                if (referenceAmount != null)
+                {
+                    if (referenceAmount < Money.Zero)
```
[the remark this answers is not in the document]
thebevrishot: class Money have a public constructor which accept long(https://github.com/MetacoSA/NBitcoin/blob/master/NBitcoin/Money.cs#L358) then a negative money can be created.

src/Ztm.Zcoin.Rpc/IZcoinRpcClient.cs
```diff
@@ -22,6 +22,8 @@ public interface IZcoinRpcClient : IDisposable
             string description,
             CancellationToken cancellationToken);
 
+
+
```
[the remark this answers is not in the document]
thebevrishot: Accidentally add. Nice catch!!

src/Ztm.Zcoin.Rpc/IZcoinRpcClient.cs
```diff
@@ -34,6 +36,9 @@ public interface IZcoinRpcClient : IDisposable
 
         Task<BitcoinAddress> GetNewAddressAsync(CancellationToken cancellationToken);
 
+        Task<(PropertyAmount Balance, PropertyAmount Reserved)> GetPropertyBalanceAsync(
```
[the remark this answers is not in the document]
thebevrishot: I think this RPC rarely change we could change it to struct once we change RPC. Named tuple is look good for me now.

src/Ztm.Zcoin.Rpc/IZcoinRpcClient.cs
```diff
@@ -55,5 +60,14 @@ public interface IZcoinRpcClient : IDisposable
             string commentTo,
             bool subtractFeeFromAmount,
             CancellationToken cancellationToken);
+
+        Task<Transaction> SendAsync(
```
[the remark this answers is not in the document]
thebevrishot: I think we need to add many exodus RPC in future and it'll be really hard to map its to known RPC name. I think we need to add Exodus prefix or Suffix as a convention for client for Exodus commands or create a separated interface for Exodus commands and implement it explicitly. How do you think?

src/Ztm.Zcoin.Rpc/ZcoinRpcClient.cs
```diff
@@ -241,6 +277,62 @@ public Task<uint256> SendRawTransactionAsync(Transaction tx, CancellationToken c
...
+                    {
+                        throw new ArgumentOutOfRangeException(nameof(referenceAmount), amount, "The value is less than zero.");
+                    }
+                    args.Add(referenceAmount.ToString());
```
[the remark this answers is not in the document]
thebevrishot: on Zcoin side it will treat it as XZC unit and try to parse it to satoshi then we can pass amount string with or without period. Then better if we specify unit here.

src/Ztm.Zcoin.Rpc.Tests/ZcoinRpcClientTests.cs
```diff
@@ -456,5 +456,226 @@ public async Task SendToAddressAsync_WithNullAmount_ShouldThrow()
                 () => this.subject.SendToAddressAsync(to, null, null, null, false, CancellationToken.None)
             );
         }
+
+        [Fact]
+        public async Task SendAsync_WithNullAsRequiredArgs_ShouldThrow()
+        {
+            var from = BitcoinAddress.Create("TDk19wPKYq91i18qmY6U9FeTdTxwPeSveo", this.node.Network);
```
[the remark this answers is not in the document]
thebevrishot: good idea.

src/Ztm.Zcoin.Rpc.Tests/ZcoinRpcClientTests.cs
```diff
@@ -456,5 +456,226 @@ public async Task SendToAddressAsync_WithNullAmount_ShouldThrow()
...
+        {
+            public BitcoinAddress owner;
+            public Ecosystem ecosystem;
+            public PropertyType propertyType;
+            public Property propertyCurrent;
+            public string catagory;
+            public string subCatagory;
+            public string name;
+            public string url;
+            public string description;
```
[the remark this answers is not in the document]
thebevrishot: Good idea

src/Ztm.Zcoin.Rpc.Tests/ZcoinRpcClientTests.cs
thebevrishot: It is good idea to change it to PropertyAmount to support both divisible and indivisible

src/Ztm.Zcoin.Rpc/IZcoinRpcClient.cs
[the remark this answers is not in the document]
ultimaweapon: If we want to go this way we need to rename `Balance` and `Reserved` to camel case to match with C# naming convention.

src/Ztm.Zcoin.Rpc/ZcoinRpcClient.cs
```diff
@@ -139,16 +132,17 @@ class BalanceDetail
                 throw new ArgumentNullException(nameof(property));
             }
 
-            var resp = await this.client.SendCommandAsync("exodus_getbalance", address.ToString(), property.Id.Value);
-            var balanceDetail = JsonConvert.DeserializeObject<BalanceDetail>(resp.ResultString);
+            var resp = await this.client.SendCommandAsync("exodus_getbalance", address.ToString(), ToNative(property.Id));
+            var rawBalance = resp.Result.Value<string>("balance");
+            var rawReserved = resp.Result.Value<string>("reserved");
 
             var balance = property.Type == PropertyType.Divisible
```
ultimaweapon: We can use `PropertyAmount.Parse` directly without checking type.

src/Ztm.Zcoin.Rpc/ZcoinRpcClient.cs
```diff
@@ -139,16 +132,17 @@ class BalanceDetail
...
-                ? PropertyAmount.FromDivisible(Convert.ToDecimal(balanceDetail.Balance))
-                : new PropertyAmount(Convert.ToInt64(balanceDetail.Balance));
+                ? PropertyAmount.FromDivisible(Convert.ToDecimal(rawBalance))
+                : new PropertyAmount(Convert.ToInt64(rawBalance));
 
             var reserved = property.Type == PropertyType.Divisible
```
ultimaweapon: Same here.

src/Ztm.Zcoin.Rpc/ZcoinRpcClient.cs
```diff
@@ -139,16 +132,17 @@ class BalanceDetail
...
-                ? PropertyAmount.FromDivisible(Convert.ToDecimal(balanceDetail.Reserved))
-                : new PropertyAmount(Convert.ToInt64(balanceDetail.Reserved));
+                ? PropertyAmount.FromDivisible(Convert.ToDecimal(rawReserved))
+                : new PropertyAmount(Convert.ToInt64(rawReserved));
 
             return (balance, reserved);
```
ultimaweapon: Better to use named field instead so it will not break when the order is changed (e.g. `(Balance: balance, Reserved: reserved)`).

src/Ztm.Zcoin.Rpc/ZcoinRpcClient.cs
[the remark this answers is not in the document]
ultimaweapon: Sorry I mean `ParseAmount` at Zcoin side.

src/Ztm.Zcoin.Rpc.Tests/ZcoinRpcClientTests.cs
```diff
@@ -558,15 +572,16 @@ public async Task GetPropertyBalanceAsync_WithValidArgs_ShouldSuccess()
...
-            var builder = new PropertyBuilder(this.node, this.subject);
-            builder.propertyType = PropertyType.Divisible;
-            var property = await builder.BuildAsync(CancellationToken.None);
+            var issuer = new PropertyIssuer(this.node, this.subject);
+            issuer.PropertyType = PropertyType.Divisible;
+            this.node.Generate(101);
+            var property = await issuer.IssueManagedAsync();
 
-            await property.GrantAsync(10_0000_0000);
+            await property.GrantAsync(new PropertyAmount(10_0000_0000));
```
ultimaweapon: We should use `PropertyAmount.FromDivisible` instead so the logic to transform value will be in a single place.

src/Ztm.Zcoin.Rpc.Tests/ZcoinRpcClientTests.cs
```diff
@@ -578,90 +593,89 @@ class ManagedProperty
...
-            public Property property { get; set; }
-
             public ManagedProperty(CoreNode node, ZcoinRpcClient client)
             {
                 this.node = node;
                 this.client = client;
             }
 
-            public async Task GrantAsync(long grantingAmount)
+            public async Task GrantAsync(PropertyAmount propertyAmount)
```
ultimaweapon: IMO just `amount` looks better.

src/Ztm.Zcoin.Rpc.Tests/ZcoinRpcClientTests.cs
```diff
@@ -578,90 +593,89 @@ class ManagedProperty
...
-            public string description;
+            public BitcoinAddress Owner { get; set; }
+            public Ecosystem Ecosystem { get; set; }
+            public PropertyType PropertyType { get; set; }
+            public Property PropertyCurrent { get; set; }
+            public string Catagory { get; set; }
+            public string SubCatagory { get; set; }
+            public string Name { get; set; }
+            public string Url { get; set; }
+            public string Description { get; set; }
```
ultimaweapon: Properties should come after constructor.

src/Ztm.Zcoin.Rpc.Tests/ZcoinRpcClientTests.cs
```diff
@@ -578,90 +593,89 @@ class ManagedProperty
...
-            public string catagory;
-            public string subCatagory;
-            public string name;
-            public string url;
-            public string description;
+            public BitcoinAddress Owner { get; set; }
+            public Ecosystem Ecosystem { get; set; }
+            public PropertyType PropertyType { get; set; }
+            public Property PropertyCurrent { get; set; }
+            public string Catagory { get; set; }
```
ultimaweapon: Seem like this is a typo.

src/Ztm.Zcoin.Rpc.Tests/ZcoinRpcClientTests.cs
```diff
@@ -578,90 +593,89 @@ class ManagedProperty
...
+            public string SubCatagory { get; set; }
```
ultimaweapon: Same here.

src/Ztm.Zcoin.Rpc.Tests/ZcoinRpcClientTests.cs
```diff
@@ -578,90 +593,89 @@ class ManagedProperty
...
-            public Property propertyCurrent;
-            public string catagory;
-            public string subCatagory;
-            public string name;
-            public string url;
-            public string description;
+            public BitcoinAddress Owner { get; set; }
+            public Ecosystem Ecosystem { get; set; }
+            public PropertyType PropertyType { get; set; }
+            public Property PropertyCurrent { get; set; }
```
ultimaweapon: IMO this looks weird. It should be `CurrentProperty`.

src/Ztm.Zcoin.Rpc/IZcoinRpcClient.cs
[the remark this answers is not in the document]
thebevrishot: I think `SendTokenAsync` look better.

src/Ztm.Zcoin.Rpc/ZcoinRpcClient.cs
thebevrishot: good idea

Issue #100, opened 2019-11-11
Attach Exodus transaction to Bitcoin transaction


ultimaweapon:
- Add extensions method for `Transaction` to retrieve/store Exodus transaction.
- Apply `ObsoleteAttribute` to the method to store Exodus transaction. We don't want it to be used everywhere else except in the specific places.
- Load Exodus transaction from Zcoin and parse it then store it to `Transaction` in every methods of `ZcoinRpcClient` that return `Transaction` both directly and indirectly.

- [x] Implementation [0.5]
- [ ] Review [1]

ultimaweapon:
Blocked by #79.

PR #111, opened 2019-11-14
Attach Exodus transaction to Bitcoin transaction

thebevrishot:
close https://github.com/zcoinofficial/ztm/issues/100


src/Ztm.Zcoin.Rpc/ZcoinRpcClient.cs
```diff
@@ -353,5 +359,73 @@ static ushort ToNative(PropertyType type)
...
+
+        async Task<TransactionInfomation> DecodeTransactionAsync(Transaction transaction)
+        {
+            var args = new List<object>()
+            {
+                transaction.ToHex(),
+            };
+
+            // Invoke RPC.
+            var resp = await this.client.SendCommandAsync("exodus_decodetransaction", args.ToArray());
```
ultimaweapon: We can eliminate `args` by replace with `transaction.ToHex()`.

ultimaweapon: I think this is a bug. If we don't provide block height to `exodus_decodetransaction` it will use current height. That mean it will always use the latest rules to validate transaction. Even with current implementation that we do this right after submitting transaction it is still a bug due to a gab between submitting and invoking `exodus_decodetransaction` still have chance to have a new block (e.g. new rule come into effect to disallow this transaction).

src/Ztm.Zcoin.NBitcoin.Tests/TransactionExtensionsTests.cs
```diff
@@ -0,0 +1,50 @@
...
+        public void GetExodusTransaction_WithUnsetValue_ShouldRetreiveNull()
+        {
+            Assert.Null(this.subject.GetExodusTransaction());
+        }
+
+        [Fact]
+        public void SetAndGetExodusTransaction_WithValidTransaction_ShouldRetreiveSameTx()
+        {
+            // Arrange.
+            var address = BitcoinAddress.Create("TG3Pnw5xPZQS8JXMVa3F9WjUFfUqXKsqAz", ZcoinNetworks.Instance.Regtest);
```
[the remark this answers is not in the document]
thebevrishot: Good idea

src/Ztm.Zcoin.Rpc/ZcoinRpcClient.cs
```diff
@@ -4,24 +4,28 @@
...
 using Ztm.Zcoin.NBitcoin.Exodus;
 
 namespace Ztm.Zcoin.Rpc
 {
     public sealed class ZcoinRpcClient : IZcoinRpcClient
     {
         readonly RPCClient client;
+        readonly ITransactionEncoder encoder;
 
-        public ZcoinRpcClient(RPCClient client)
```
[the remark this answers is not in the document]
thebevrishot: nice catch

src/Ztm.Zcoin.Rpc/ZcoinRpcClient.cs
thebevrishot: Removed, Use `exodus_gettransaction` instead.

src/Ztm.Zcoin.Rpc/ZcoinRpcClient.cs
```diff
@@ -362,70 +388,83 @@ static ushort ToNative(PropertyType type)
 
         class TransactionInfomation
```
ultimaweapon: Once we move `GetExodusTransactionAsync` to public we need to move this one to a dedicated file.

ultimaweapon: `TransactionInfomation` is ambiguous. IMO rename to `ExodusTransactionInformation`.

src/Ztm.Zcoin.Rpc/ZcoinRpcClient.cs
```diff
@@ -362,70 +388,83 @@ static ushort ToNative(PropertyType type)
...
+                Valid = resp.Result.Value<bool>("valid"),
+                InvalidReason = resp.Result.Value<string>("invalidreason"),
                 Version = resp.Result.Value<int>("version"),
                 TypeInt = resp.Result.Value<int>("type_int"),
                 Type = resp.Result.Value<string>("type")
             };
         }
 
-        async Task<(string Payload, int PayloadSize)> GetPayloadAsync(Transaction transaction)
+        async Task<(string payload, int payloadSize)> GetPayloadAsync(Transaction transaction)
```
ultimaweapon: Same here. We should move this one to public too.

ultimaweapon: Once we move this to public we should accept `uint256` instead.

ultimaweapon: `GetPayloadAsync` is ambiguous. IMO rename to `GetExodusPayload`.

src/Ztm.Zcoin.Rpc/ZcoinRpcClient.cs
```diff
@@ -362,70 +388,83 @@ static ushort ToNative(PropertyType type)
...
         {
             var resp = await this.client.SendCommandAsync("exodus_getpayload", new object[]{transaction.GetHash()});
```
ultimaweapon: We don't need to construct an array to pass RPC argument due to `SendCommandAsync` accept variadic parameters.

src/Ztm.Zcoin.Rpc/ZcoinRpcClient.cs
ultimaweapon: We should return POCO instead once we moved to public.

src/Ztm.Zcoin.Rpc/ZcoinRpcClient.cs
ultimaweapon: We should return `null` instead if transaction is not Exodus transaction.

src/Ztm.Zcoin.Rpc/ZcoinRpcClient.cs
```diff
@@ -362,70 +388,83 @@ static ushort ToNative(PropertyType type)
...
 
-            ExodusTransaction tx;
+            var payload = await GetPayloadAsync(transaction);
 
             try
             {
-                tx = this.encoder.Decode(infomation.SendingAddress, infomation.ReferenceAddress, Encoders.Hex.DecodeData(payload.Payload));
+                return this.encoder.Decode(infomation.SendingAddress, infomation.ReferenceAddress, Encoders.Hex.DecodeData(payload.payload));
             }
             catch (TransactionFieldException)
```
ultimaweapon: The decode should never fail. If it fail that mean we not follow consensus. So, we should not catch this error.

src/Ztm.Zcoin.Rpc/ZcoinRpcClient.cs
```diff
@@ -4,24 +4,33 @@
...
 using Newtonsoft.Json.Linq;
+using Ztm.Zcoin.NBitcoin;
 using Ztm.Zcoin.NBitcoin.Exodus;
 
 namespace Ztm.Zcoin.Rpc
 {
     public sealed class ZcoinRpcClient : IZcoinRpcClient
     {
         readonly RPCClient client;
+        readonly ITransactionEncoder encoder;
```
ultimaweapon: `encoder` is quite ambiguous. IMO rename to `exodusEncoder`.

src/Ztm.Zcoin.Rpc/ZcoinRpcClientFactory.cs
```diff
@@ -12,8 +13,9 @@ public sealed class ZcoinRpcClientFactory : IZcoinRpcClientFactory
         readonly Uri serverUri;
         readonly NetworkType networkType;
         readonly RPCCredentialString credential;
+        readonly ITransactionEncoder encoder;
```
ultimaweapon: Same here.

src/Ztm.WebApi/ServicesExtensions.cs
```diff
@@ -0,0 +1,17 @@
+using Microsoft.Extensions.DependencyInjection;
+using Ztm.Zcoin.NBitcoin.Exodus;
+
+namespace Ztm.WebApi
+{
+    public static class ServicesExtensions
```
ultimaweapon: This one should go to `Ztm.Zcoin.NBitcoin` instead.

ultimaweapon: The name of extension class should start with what it extend. So in this case should be `ServiceCollectionExtensions`.

src/Ztm.WebApi/ServicesExtensions.cs
```diff
@@ -0,0 +1,17 @@
...
+    {
+        public static void AddTransactionEncoder(this IServiceCollection services)
```
ultimaweapon: Once we moved this file we should rename this to `AddNBitcoin`.

src/Ztm.WebApi/ServicesExtensions.cs
```diff
@@ -0,0 +1,17 @@
...
+        {
+            // Payload Encoders
+            services.AddTransient<ITransactionPayloadEncoder, SimpleSendEncoder>();
+
+            // Encoder
+            services.AddTransient<ITransactionEncoder, TransactionEncoder>();
```
ultimaweapon: This one should be singleton.

src/Ztm.WebApi/ServicesExtensions.cs
```diff
@@ -0,0 +1,17 @@
...
+using Ztm.Zcoin.NBitcoin.Exodus;
+
+namespace Ztm.WebApi
+{
+    public static class ServicesExtensions
+    {
+        public static void AddTransactionEncoder(this IServiceCollection services)
+        {
+            // Payload Encoders
+            services.AddTransient<ITransactionPayloadEncoder, SimpleSendEncoder>();
```
ultimaweapon: Same here.

src/Ztm.WebApi/Startup.cs
```diff
@@ -46,6 +47,9 @@ public void ConfigureServices(IServiceCollection services)
             services.AddTransient<IBlocksRetriever, BlocksRetriever>();
             services.AddSingleton<IBlocksStorage, BlocksStorage>();
 
+            // Transaction Encoder Services.
```
ultimaweapon: We need to update this comment once we rename the method.

src/Ztm.WebApi/Startup.cs
```diff
@@ -86,7 +90,8 @@ IZcoinRpcClientFactory CreateZcoinRpcClientFactory(IServiceProvider provider)
             return new ZcoinRpcClientFactory(
                 config.Rpc.Address,
                 config.Network.Type,
-                RPCCredentialString.Parse($"{config.Rpc.UserName}:{config.Rpc.Password}")
+                RPCCredentialString.Parse($"{config.Rpc.UserName}:{config.Rpc.Password}"),
+                (ITransactionEncoder)provider.GetRequiredService(typeof(ITransactionEncoder))
```
ultimaweapon: We can use `provider.GetRequiredService<ITransactionEncoder>()` to remove casting.

src/Ztm.Zcoin.Rpc/ZcoinRpcClient.cs
```diff
@@ -362,70 +388,83 @@ static ushort ToNative(PropertyType type)
...
             return new TransactionInfomation{
                 TxId = uint256.Parse(resp.Result.Value<string>("txid")),
-                Fee = Money.Parse(resp.Result.Value<string>("fee")),
                 SendingAddress = BitcoinAddress.Create(resp.Result.Value<string>("sendingaddress"), this.client.Network),
                 ReferenceAddress = (rawRefAddress == null || rawRefAddress == string.Empty)
                     ? null : BitcoinAddress.Create(resp.Result.Value<string>("referenceaddress"), this.client.Network),
                 IsMine = resp.Result.Value<bool>("ismine"),
+                Confirmations = resp.Result.Value<int>("confirmations"),
+                Fee = Money.Parse(resp.Result.Value<string>("fee")),
+                BlockTime = resp.Result.Value<int>("blocktime"),
```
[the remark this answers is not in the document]
thebevrishot: Use `DateTimeOffset` instead because NBitcoin use this type.

src/Ztm.Zcoin.Rpc/ZcoinRpcClient.cs
```diff
@@ -362,70 +388,83 @@ static ushort ToNative(PropertyType type)
...
             // Invoke RPC.
-            var resp = await this.client.SendCommandAsync("exodus_decodetransaction", args.ToArray());
+            var resp = await this.client.SendCommandAsync("exodus_gettransaction", transaction.GetHash());
             var rawRefAddress = resp.Result.Value<string>("referenceaddress");
 
             return new TransactionInfomation{
                 TxId = uint256.Parse(resp.Result.Value<string>("txid")),
-                Fee = Money.Parse(resp.Result.Value<string>("fee")),
                 SendingAddress = BitcoinAddress.Create(resp.Result.Value<string>("sendingaddress"), this.client.Network),
                 ReferenceAddress = (rawRefAddress == null || rawRefAddress == string.Empty)
```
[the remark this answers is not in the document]
thebevrishot: Good idea!!

src/Ztm.Zcoin.Rpc/ZcoinRpcClient.cs
```diff
@@ -362,70 +388,83 @@ static ushort ToNative(PropertyType type)
...
                     ? null : BitcoinAddress.Create(resp.Result.Value<string>("referenceaddress"), this.client.Network),
```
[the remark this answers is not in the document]
thebevrishot: Accidentally copy. NICE CATCH!!

src/Ztm.Zcoin.Rpc/ZcoinRpcClient.cs
thebevrishot: `GetExodusPayloadAsync` should be ok.

thebevrishot: return only `byte[]` instead

src/Ztm.Zcoin.Rpc/ZcoinRpcClient.cs
thebevrishot: just let it throw for now, already create an issue.

src/Ztm.Zcoin.Rpc/ZcoinRpcClient.cs
thebevrishot: nice catch

src/Ztm.WebApi/ServicesExtensions.cs
thebevrishot: moved

src/Ztm.Zcoin.Rpc/IZcoinRpcClient.cs
```diff
@@ -32,8 +32,12 @@ public interface IZcoinRpcClient : IDisposable
 
         Task<BlockchainInfo> GetBlockchainInfoAsync(CancellationToken cancellationToken);
 
+        Task<ExodusTransactionInformation> GetExodusTransactionAsync(uint256 transaction, CancellationToken cancellationToken);
+
         Task<BitcoinAddress> GetNewAddressAsync(CancellationToken cancellationToken);
 
+        Task<byte[]> GetExodusPayloadAsync(uint256 transaction, CancellationToken cancellationToken);
```
ultimaweapon: This one should come before `GetExodusTransactionAsync`.

src/Ztm.Zcoin.Rpc/ZcoinRpcClient.cs
```diff
@@ -386,73 +409,27 @@ static ushort ToNative(PropertyType type)
...
-            var payload = await GetPayloadAsync(transaction);
+            var payload = await GetExodusPayloadAsync(transaction.GetHash(), CancellationToken.None);
 
             try
             {
-                return this.encoder.Decode(infomation.SendingAddress, infomation.ReferenceAddress, Encoders.Hex.DecodeData(payload.payload));
+                return this.exodusEncoder.Decode(infomation.SendingAddress, infomation.ReferenceAddress, payload);
             }
-            catch (TransactionFieldException)
+            catch (TransactionFieldException ex)
```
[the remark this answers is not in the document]
thebevrishot: good idea

src/Ztm.Zcoin.Rpc/ZcoinRpcClient.cs
```diff
@@ -411,25 +421,26 @@ static ushort ToNative(PropertyType type)
...
         async Task<ExodusTransaction> TryDecodeExodusTransaction(Transaction transaction)
         {
-            var infomation = await GetExodusTransactionAsync(transaction.GetHash(), CancellationToken.None);
-            if (infomation == null)
+            ExodusTransactionInformation infomation;
+            try
+            {
+                infomation = await GetExodusTransactionAsync(transaction.GetHash(), CancellationToken.None);
+            }
+            catch (Exception) // lgtm [cs/catch-of-all-exceptions]
```
ultimaweapon: This is dangerous due to it will also include other errors than transaction is not Exodus transaction.

src/Ztm.Zcoin.Rpc/ZcoinRpcClient.cs
```diff
@@ -426,7 +426,7 @@ async Task<ExodusTransaction> TryDecodeExodusTransaction(Transaction transaction
             {
                 infomation = await GetExodusTransactionAsync(transaction.GetHash(), CancellationToken.None);
             }
-            catch (Exception) // lgtm [cs/catch-of-all-exceptions]
+            catch (RPCException)
```
ultimaweapon: This still dangerous due to it cannot distinguished what is the error. We should catch only if it is non-Exodus transaction error.

src/Ztm.Zcoin.Rpc.Tests/ZcoinRpcClientTests.cs
```diff
@@ -712,6 +714,92 @@ public async Task GetExodusPayloadAsync_WithValidExistTransaction_ShouldSuccess(
...
+            await property.GrantAsync(new PropertyAmount(1000));
+
+            var destination = await this.subject.GetNewAddressAsync(CancellationToken.None);
+            var rawTx = await this.subject.SendTokenAsync(property.Owner, destination, property.Property, new PropertyAmount(10), null, null, CancellationToken.None);
+
+            await this.subject.SendRawTransactionAsync(rawTx, CancellationToken.None);
+            var id = this.node.Generate(1).First();
+
+            var exodusTransaction = new TestExodusTransaction(TestAddress.Regtest1, TestAddress.Regtest2);
+            this.encoder.Decode(Arg.Any<BitcoinAddress>(), Arg.Any<BitcoinAddress>(), Arg.Any<byte[]>())
```
ultimaweapon: Better to match with sender and receiver so at least we know it is actually what we expect.

src/Ztm.Zcoin.Rpc.Tests/ZcoinRpcClientTests.cs
```diff
@@ -712,6 +714,92 @@ public async Task GetExodusPayloadAsync_WithValidExistTransaction_ShouldSuccess(
...
+            var id = this.node.Generate(1).First();
+            var block = await this.subject.GetBlockAsync(id, CancellationToken.None);
+
+            await Assert.ThrowsAsync<RPCException>(
+                () => this.subject.GetExodusTransactionAsync(block.Transactions.First().GetHash(), CancellationToken.None)
+            );
+        }
+
+        [Fact]
+        public async Task GetExodusTransactionAsync_InExistTx_ShouldThrow()
```
ultimaweapon: I think `InExistTx` is quite confuse.

src/Ztm.Zcoin.Rpc.Tests/ZcoinRpcClientFactoryTests.cs
```diff
@@ -3,19 +3,28 @@
...
+using NSubstitute;
 using Xunit;
+using Ztm.Zcoin.NBitcoin.Exodus;
 using Ztm.Zcoin.Testing;
 
 namespace Ztm.Zcoin.Rpc.Tests
 {
     public class ZcoinRpcClientFactoryTests
     {
+        readonly ITransactionEncoder encoder;
```
ultimaweapon: Same as previous issue.

Issue #62, opened 2019-10-21
Implement generic address pool


ultimaweapon:
- Have method to generate, lock and release address.
- The lock method will find address that not locked and mark it lock then return that address. If no available address it will return null.
- The release method will unlock the specified address.
- The data in the pool will be handled by external service.

- [x] Implementation [1]
- [ ] Review [0.5]

PR #121, opened 2019-11-25
Address pool

thebevrishot:
closes https://github.com/zcoinofficial/ztm/issues/62


src/Ztm.WebApi/AddressPools/ReceivingAddressPool.cs
```diff
@@ -0,0 +1,120 @@
...
+        public async Task<ReceivingAddressReservation> TryLockAddressAsync(CancellationToken cancellationToken)
+        {
+            this.storageLock.EnterWriteLock();
+
+            try
+            {
+                var addresses = await this.storage.ListReceivingAddressAsync(cancellationToken);
+                if (addresses != null)
+                {
+                    var availables = addresses.Where(a => a.Available);
```
ultimaweapon: The filtering should be done on storage side to be more efficient.

src/Ztm.WebApi/AddressPools/ReceivingAddressPool.cs
```diff
@@ -0,0 +1,120 @@
...
+
+                    if (availables.Count() <= 0)
+                    {
+                        return null;
+                    }
+
+                    var chosen = this.choser.Choose(availables);
+
+                    if (chosen != null)
```
ultimaweapon: The choosing algorithm should never return `null` if there is at least one item to choose.

src/Ztm.WebApi/AddressPools/ReceivingAddressPool.cs
```diff
@@ -0,0 +1,120 @@
...
+                    {
+                        var reservation = await this.storage.CreateReservationAsync(chosen.Id, cancellationToken);
+                        await this.storage.SetLockedStatusAsync(chosen.Id, true, cancellationToken);
```
ultimaweapon: The locked state should be combined together with reservation creation and must be atomic operation.

src/Ztm.WebApi/AddressPools/ReceivingAddressPool.cs
```diff
@@ -0,0 +1,120 @@
...
+                    throw new KeyNotFoundException("The reservation is not found.");
+                }
+
+                if (r.ReleasedDate != null)
+                {
+                    throw new InvalidOperationException("The reservation is already released.");
+                }
+
+                await this.storage.SetLockedStatusAsync(r.ReceivingAddress.Id, false, cancellationToken);
+                await this.storage.SetReleasedTimeAsync(id, cancellationToken);
```
ultimaweapon: The above two operations should be combine together and must be atomic operation.

src/Ztm.WebApi/AddressPools/ReceivingAddress.cs
```diff
@@ -0,0 +1,45 @@
...
+using NBitcoin;
+
+namespace Ztm.WebApi.AddressPools
+{
+    public sealed class ReceivingAddress
+    {
+        public Guid Id { get; }
+        public BitcoinAddress Address { get; }
+        public bool IsLocked { get; }
+        public List<ReceivingAddressReservation> ReceivingAddressReservations { get; }
```
ultimaweapon: Properties should come after constructor.

ultimaweapon: I think just `Reservations` is enough due to we already have context.

ultimaweapon: Better to use `IEnumerable<ReceivingAddressReservation>` if mutable is not required; otherwise better to use `ICollection<ReceivingAddressReservation>` instead.

src/Ztm.WebApi/AddressPools/ReceivingAddress.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        public ReceivingAddress(Guid id, BitcoinAddress address, bool isLocked, List<ReceivingAddressReservation> receivingAddressReservations)
```
ultimaweapon: Same here.

ultimaweapon: We should not allow `address` to be `null`.

ultimaweapon: We should not allow `receivingAddressReservations` to be `null`.

src/Ztm.WebApi/AddressPools/ReceivingAddress.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            this.Id = id;
+            this.Address = address;
+            this.IsLocked = isLocked;
+            this.ReceivingAddressReservations = receivingAddressReservations;
+        }
+
+        public bool Available
```
ultimaweapon: We should remove this property. The problem is it will cause more way to check if it available or not, which is not consistent.

src/Ztm.WebApi/AddressPools/IAddressChoser.cs
```diff
@@ -0,0 +1,11 @@
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Ztm.WebApi.AddressPools
+{
+    public interface IAddressChoser
+    {
+        // Garantee all receiving addresses that are send to this function will be available and size of set is not be null.
```
ultimaweapon: We should use documentation tag instead of normal comment.

src/Ztm.WebApi/AddressPools/IAddressChoser.cs
```diff
@@ -0,0 +1,11 @@
...
+        ReceivingAddress Choose(IEnumerable<ReceivingAddress> receivingAddress);
```
ultimaweapon: I think just `addresses` should be enough.

src/Ztm.WebApi/AddressPools/LessUsageFirstChoser.cs
```diff
@@ -0,0 +1,31 @@
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Ztm.WebApi.AddressPools
+{
+    public sealed class LessUsageFirstChoser : IAddressChoser
+    {
+        public ReceivingAddress Choose(IEnumerable<ReceivingAddress> receivingAddress)
```
ultimaweapon: We should throw if `receivingAddress` is `null`.

src/Ztm.WebApi/AddressPools/LessUsageFirstChoser.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            var chosen = receivingAddress.First();
+            var currentUsage = chosen.ReceivingAddressReservations == null
```
ultimaweapon: `ReceivingAddressReservations` should never be `null`.

src/Ztm.WebApi/AddressPools/LessUsageFirstChoser.cs
```diff
@@ -0,0 +1,31 @@
...
+                ? 0
+                : chosen.ReceivingAddressReservations.Count;
+
+            foreach (var candidate in receivingAddress)
+            {
+                var usage = candidate.ReceivingAddressReservations == null
```
ultimaweapon: Same here.

src/Ztm.WebApi/AddressPools/LessUsageFirstChoser.cs
ultimaweapon: We should throw if `receivingAddress` is empty.

ultimaweapon: Seems like all of logic in this method can be replaced by one line

```csharp
return receivingAddress.Aggregate((previous, next) => next.ReceivingAddressReservations.Count < previous.ReceivingAddressReservations.Count ? next : previous);
```

src/Ztm.WebApi/AddressPools/ReceivingAddressPool.cs
```diff
@@ -0,0 +1,120 @@
...
+
+namespace Ztm.WebApi.AddressPools
+{
+    public sealed class ReceivingAddressPool : IReceivingAddressPool
+    {
+        readonly IAddressGenerator generator;
+        readonly IReceivingAddressStorage storage;
+        readonly IAddressChoser choser;
+
+        readonly ReaderWriterLockSlim storageLock;
```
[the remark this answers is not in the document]
thebevrishot: Firstly, I decide to move logic to pool because need to simplify storage. However, we need to support case that addresses in storage is shares between other user. I decide to move the logic back to storage by reverting a commit.

src/Ztm.WebApi/AddressPools/ReceivingAddressPool.cs
thebevrishot: Add filter flags

src/Ztm.WebApi/AddressPools/ReceivingAddressPool.cs
```diff
@@ -37,81 +34,27 @@ public sealed class ReceivingAddressPool : IReceivingAddressPool
...
-            try
-            {
                 var address = await this.generator.GenerateAsync(cancellationToken);
-                await this.storage.AddAddressAsync(address, cancellationToken);
-            }
-            finally
-            {
-                storageLock.ExitWriteLock();
-            }
+                await this.storage.AddAddressAsync(address, CancellationToken.None);
```
ultimaweapon: Seems like 2 lines above have indentation error.

src/Ztm.WebApi/AddressPools/ReceivingAddressPool.cs
```diff
@@ -37,81 +34,27 @@ public sealed class ReceivingAddressPool : IReceivingAddressPool
...
-                }
-
-                await this.storage.SetLockedStatusAsync(r.ReceivingAddress.Id, false, cancellationToken);
-                await this.storage.SetReleasedTimeAsync(id, cancellationToken);
-            }
-            finally
-            {
-                this.storageLock.ExitWriteLock();
-            }
+            return this.ReleaseAddressAsync(id, cancellationToken);
```
ultimaweapon: This will cause stack overflow! :bomb:

src/Ztm.WebApi/AddressPools/ReceivingAddressPool.cs
```diff
@@ -37,81 +34,27 @@ public sealed class ReceivingAddressPool : IReceivingAddressPool
...
-                        return reservation;
-                    }
-                }
-            }
-            finally
+            if (addresses.Any())
             {
-                this.storageLock.ExitWriteLock();
+                var chosen = this.choser.Choose(addresses);
+                return await this.storage.TryLockAsync(chosen.Id, CancellationToken.None);
```
ultimaweapon: We can allow to cancel here due to all of above operations does not cause any state to change.

src/Ztm.WebApi/AddressPools/LessUsageFirstChoser.cs
```diff
@@ -1,31 +1,29 @@
...
+            }
 
-                if (usage < currentUsage)
-                {
-                    chosen = candidate;
-                    currentUsage = usage;
-                }
+            if (!addresses.Any())
+            {
+                throw new ArgumentOutOfRangeException(nameof(addresses), "Addresses could not be empty.");
```
ultimaweapon: I think `ArgumentOutOfRangeException` is not quite correct here. We should use `ArgumentException` instead.

src/Ztm.WebApi/AddressPools/LessUsageFirstChoser.cs
```diff
@@ -1,31 +1,29 @@
...
             }
 
-            return chosen;
+            return addresses.Aggregate
+            (
+                (previous, next) => next.Reservations.Count < previous.Reservations.Count
+                ? next
+                : previous
```
ultimaweapon: The indentation for 2 lines above should not be the same as line 23 due to it is the same statement.

src/Ztm.WebApi/AddressPools/SqlReceivingAddressStorage.cs
```diff
@@ -0,0 +1,202 @@
...
+                }
+
+                return Task.FromResult<IEnumerable<ReceivingAddress>>(query.Select(r => ToDomain(r)).ToList());
+            }
+        }
+
+        public async Task ReleaseAsync(Guid id, CancellationToken cancellationToken)
+        {
+            using (var db = this.databaseFactory.CreateDbContext())
+            using (var tx = db.Database.BeginTransaction())
```
ultimaweapon: The default isolation level is quite dangerous for operations we are going to do, which usually is Read Committed. Better to use Repeatable Read instead.

src/Ztm.WebApi/AddressPools/SqlReceivingAddressStorage.cs
```diff
@@ -0,0 +1,202 @@
...
+            {
+                var reservation = await db.ReceivingAddressReservations
+                    .Include(r => r.ReceivingAddress)
+                    .FirstAsync(r => r.Id == id);
```
ultimaweapon: This will cause thread to block.

src/Ztm.WebApi/AddressPools/SqlReceivingAddressStorage.cs
```diff
@@ -0,0 +1,202 @@
...
+
+                if (reservation.ReleasedAt != null)
+                {
+                    throw new InvalidOperationException("The reservation is already released.");
+                }
+
+                reservation.ReleasedAt = DateTime.UtcNow;
+                reservation.ReceivingAddress.IsLocked = false;
+
+                await db.SaveChangesAsync();
```
ultimaweapon: We should allow to cancel here due to all of above operations did not change state.

src/Ztm.WebApi/AddressPools/SqlReceivingAddressStorage.cs
```diff
@@ -0,0 +1,202 @@
...
+                tx.Commit();
+            }
+        }
+
+        public async Task<ReceivingAddressReservation> TryLockAsync(Guid id, CancellationToken cancellationToken)
+        {
+            using (var db = this.databaseFactory.CreateDbContext())
+            using (var tx = db.Database.BeginTransaction())
```
ultimaweapon: Same as previous issue.

src/Ztm.WebApi/AddressPools/SqlReceivingAddressStorage.cs
```diff
@@ -0,0 +1,202 @@
...
+            {
+                var recv = await db.ReceivingAddresses.FirstAsync(a => a.Id == id, cancellationToken);
```
ultimaweapon: Same as previous issue.

src/Ztm.WebApi/AddressPools/SqlReceivingAddressStorage.cs
```diff
@@ -0,0 +1,202 @@
...
+                if (recv == null)
+                {
+                    return null;
+                }
+
+                if (recv.IsLocked)
+                {
+                    return null;
```
ultimaweapon: We should throw instead if it is locked so the caller can different between not found and locked.

src/Ztm.WebApi/AddressPools/SqlReceivingAddressStorage.cs
```diff
@@ -0,0 +1,202 @@
...
+                }
+
+                recv.IsLocked = true;
+
+                var lockedAt = DateTime.UtcNow;
+                var reservation = await db.AddAsync
```
ultimaweapon: Same as previous issue.

src/Ztm.WebApi/AddressPools/SqlReceivingAddressStorage.cs
```diff
@@ -0,0 +1,202 @@
...
+                }
+
+                if (recv.IsLocked)
+                {
+                    return null;
+                }
+
+                recv.IsLocked = true;
+
+                var lockedAt = DateTime.UtcNow;
```
ultimaweapon: `lockedAt` used only once so better to inline instead.

src/Ztm.WebApi/AddressPools/SqlReceivingAddressStorage.cs
```diff
@@ -0,0 +1,202 @@
...
+                var reservation = await db.AddAsync
+                (
+                    new ReceivingAddressReservationModel{
```
ultimaweapon: Brace error.

src/Ztm.WebApi/AddressPools/SqlReceivingAddressStorage.cs
```diff
@@ -0,0 +1,202 @@
...
+                );
+
+                await db.SaveChangesAsync(cancellationToken);
+                tx.Commit();
+
+                return ToDomain(reservation.Entity);
+            }
+        }
+
+        ReceivingAddress ToDomain(ReceivingAddressModel receivingAddress)
```
ultimaweapon: Rename `receivingAddress` to `entity` should be better.

src/Ztm.WebApi/AddressPools/SqlReceivingAddressStorage.cs
```diff
@@ -0,0 +1,202 @@
...
+        {
+            var r = new ReceivingAddress
+            (
+                receivingAddress.Id,
+                BitcoinAddress.Create(receivingAddress.Address, this.network),
+                receivingAddress.IsLocked,
+                new List<ReceivingAddressReservation>()
+            );
+
+            if (receivingAddress.ReceivingAddressReservations != null)
```
ultimaweapon: `ReceivingAddressReservations` should never be `null`.

src/Ztm.WebApi/AddressPools/SqlReceivingAddressStorage.cs
```diff
@@ -0,0 +1,202 @@
...
+                foreach (var reservation in receivingAddress.ReceivingAddressReservations)
+                {
+                    r.Reservations.Add(ToDomain(reservation, r));
+                }
+            }
+
+            return r;
+        }
+
+        ReceivingAddressReservation ToDomain(ReceivingAddressReservationModel reservationModel)
```
ultimaweapon: Same here.

src/Ztm.WebApi/AddressPools/SqlReceivingAddressStorage.cs
```diff
@@ -0,0 +1,202 @@
...
+                reservationModel.Id,
+                ToDomain(reservationModel.ReceivingAddress),
+                DateTime.SpecifyKind(reservationModel.LockedAt, DateTimeKind.Utc),
+                reservationModel.ReleasedAt.HasValue
+                    ? DateTime.SpecifyKind(reservationModel.ReleasedAt.Value, DateTimeKind.Utc)
+                    : new Nullable<DateTime>()
+            );
+        }
+
+        ReceivingAddressReservation ToDomain(ReceivingAddressReservationModel reservationModel, ReceivingAddress address)
```
ultimaweapon: Same here.

ultimaweapon: This method and the above one almost identical. Better to merge it together.

src/Ztm.Data.Entity/Contexts/Main/ReceivingAddressReservation.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+
+namespace Ztm.Data.Entity.Contexts.Main
+{
+    public sealed class ReceivingAddressReservation
+    {
+        public Guid Id { get; set; }
+        public Guid ReceivingAddressId { get; set; }
```
ultimaweapon: Just `AddressId` should be enough due to we already have context.

src/Ztm.Data.Entity/Contexts/Main/ReceivingAddressReservation.cs
```diff
@@ -0,0 +1,14 @@
...
+        public DateTime LockedAt { get; set; }
+        public DateTime? ReleasedAt { get; set; }
+
+        public ReceivingAddress ReceivingAddress { get; set; }
```
ultimaweapon: Same here.

src/Ztm.Data.Entity/Contexts/Main/ReceivingAddress.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Ztm.Data.Entity.Contexts.Main
+{
+    public sealed class ReceivingAddress
+    {
+        public Guid Id { get; set; }
+        public string Address { get; set; }
+        public bool IsLocked { get; set; }
+
+        public List<ReceivingAddressReservation> ReceivingAddressReservations { get; set; }
```
ultimaweapon: Just `Reservations` looks better.

src/Ztm.Data.Entity/Contexts/MainDatabase.cs
```diff
@@ -30,6 +30,10 @@ public MainDatabase(DbContextOptions<MainDatabase> options) : base(options)
 
         public DbSet<WebApiCallbackHistory> WebApiCallbackHistories { get; set; }
 
+        public DbSet<ReceivingAddress> ReceivingAddresses { get; set; }
+
+        public DbSet<ReceivingAddressReservation> ReceivingAddressReservations { get; set; }
```
ultimaweapon: The above properties should ordered in ascending.

src/Ztm.Data.Entity/Contexts/MainDatabase.cs
```diff
@@ -138,6 +142,30 @@ protected virtual void ConfigureWebApiCallbackHistory(EntityTypeBuilder<WebApiCa
                    .OnDelete(DeleteBehavior.Cascade);
         }
 
+        protected virtual void ConfigureReceivingAddress(EntityTypeBuilder<ReceivingAddress> builder)
+        {
+            builder.Property(e => e.Id).IsRequired().ValueGeneratedOnAdd();
+            builder.Property(e => e.Address).IsRequired();
+            builder.Property(e => e.IsLocked).HasDefaultValue(false);
```
ultimaweapon: We should apply `IsRequired` too so it match with the others.

src/Ztm.Data.Entity/Contexts/MainDatabase.cs
```diff
@@ -138,6 +142,30 @@ protected virtual void ConfigureWebApiCallbackHistory(EntityTypeBuilder<WebApiCa
...
+
+            builder.HasKey(e => e.Id);
+            builder.HasIndex(e => e.Address).IsUnique();
+        }
+
+        protected virtual void ConfigureReceivingAddressReservation(EntityTypeBuilder<ReceivingAddressReservation> builder)
+        {
+            builder.Property(e => e.Id).IsRequired().ValueGeneratedOnAdd();
+            builder.Property(e => e.ReceivingAddressId).IsRequired();
+            builder.Property(e => e.LockedAt).IsRequired();
```
ultimaweapon: We need to add `ReleasedAt` too even we have nothing to configure so it will be more easy to know how much columns it have.

src/Ztm.Data.Entity/Contexts/MainDatabase.cs
```diff
@@ -138,6 +142,30 @@ protected virtual void ConfigureWebApiCallbackHistory(EntityTypeBuilder<WebApiCa
...
+
+            builder.HasKey(e => e.Id);
```
ultimaweapon: We need to add key for `ReceivingAddressId` too even it is implicit created so it is more easy to read.

src/Ztm.Data.Entity/Contexts/MainDatabase.cs
```diff
@@ -138,6 +142,30 @@ protected virtual void ConfigureWebApiCallbackHistory(EntityTypeBuilder<WebApiCa
                    .OnDelete(DeleteBehavior.Cascade);
         }
 
+        protected virtual void ConfigureReceivingAddress(EntityTypeBuilder<ReceivingAddress> builder)
```
ultimaweapon: Protected method should be ordered in ascending.

src/Ztm.Data.Entity/Contexts/MainDatabase.cs
```diff
@@ -138,6 +142,30 @@ protected virtual void ConfigureWebApiCallbackHistory(EntityTypeBuilder<WebApiCa
...
+        {
+            builder.Property(e => e.Id).IsRequired().ValueGeneratedOnAdd();
+            builder.Property(e => e.Address).IsRequired();
+            builder.Property(e => e.IsLocked).HasDefaultValue(false);
+
+            builder.HasKey(e => e.Id);
+            builder.HasIndex(e => e.Address).IsUnique();
+        }
+
+        protected virtual void ConfigureReceivingAddressReservation(EntityTypeBuilder<ReceivingAddressReservation> builder)
```
ultimaweapon: Same here.

src/Ztm.Data.Entity/Contexts/MainDatabase.cs
```diff
@@ -147,6 +175,8 @@ protected sealed override void OnModelCreating(ModelBuilder modelBuilder)
             modelBuilder.Entity<Transaction>(ConfigureTransaction);
             modelBuilder.Entity<WebApiCallback>(ConfigureWebApiCallback);
             modelBuilder.Entity<WebApiCallbackHistory>(ConfigureWebApiCallbackHistory);
+            modelBuilder.Entity<ReceivingAddress>(ConfigureReceivingAddress);
+            modelBuilder.Entity<ReceivingAddressReservation>(ConfigureReceivingAddressReservation);
```
ultimaweapon: Better to order by ascending too so it match with the method order.

src/Ztm.WebApi/AddressPools/IReceivingAddressStorage.cs
```diff
@@ -0,0 +1,19 @@
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using NBitcoin;
+
+namespace Ztm.WebApi.AddressPools
+{
+    public interface IReceivingAddressStorage
```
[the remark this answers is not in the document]
thebevrishot: Make sense

src/Ztm.WebApi/AddressPools/SqlReceivingAddressStorage.cs
```diff
@@ -0,0 +1,202 @@
...
+                {
+                    query = query.Where(a => !a.IsLocked);
+                }
+
+                if (filter.HasFlag(AddressFilter.NeverUsed))
+                {
+                    query = query.Where(a => !a.ReceivingAddressReservations.Any());
+                }
+
+                return Task.FromResult<IEnumerable<ReceivingAddress>>(query.Select(r => ToDomain(r)).ToList());
```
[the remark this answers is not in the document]
thebevrishot: Nice catch

src/Ztm.WebApi/AddressPools/SqlReceivingAddressStorage.cs
thebevrishot: Seem like it's non-blocking.

src/Ztm.WebApi/AddressPools/IReceivingAddressStorage.cs
```diff
@@ -8,7 +8,7 @@ namespace Ztm.WebApi.AddressPools
 {
     public interface IReceivingAddressStorage
     {
-        Task<ReceivingAddress> AddAddressAsync(BitcoinAddress address, CancellationToken cancellationToken);
+        Task<ReceivingAddress> AddAsync(BitcoinAddress address, CancellationToken cancellationToken);
         Task<ReceivingAddress> GetAsync(Guid id, CancellationToken cancellationToken);
         Task<IEnumerable<ReceivingAddress>> ListReceivingAddressAsync(
```
ultimaweapon: We forgot to rename this one.

src/Ztm.WebApi/AddressPools/EntityReceivingAddressStorage.cs
```diff
@@ -53,6 +53,7 @@ public async Task<ReceivingAddress> AddAddressAsync(BitcoinAddress address, Canc
 
                 await db.SaveChangesAsync(cancellationToken);
 
+                recAddress.Entity.Reservations = new List<ReceivingAddressReservationModel>();
```
ultimaweapon: We should handle this in the constructor of the model instead.

src/Ztm.WebApi/AddressPools/EntityReceivingAddressStorage.cs
```diff
@@ -62,18 +63,18 @@ public async Task<ReceivingAddress> GetAsync(Guid id, CancellationToken cancella
             using (var db = this.databaseFactory.CreateDbContext())
             {
                 var recv = await db.ReceivingAddresses
```
ultimaweapon: I think we should return `null` if no matching address.

src/Ztm.WebApi/AddressPools/EntityReceivingAddressStorage.cs
```diff
@@ -82,21 +83,21 @@ public Task<IEnumerable<ReceivingAddress>> ListReceivingAddressAsync(AddressFilt
...
-            using (var tx = db.Database.BeginTransaction())
+            using (var tx = db.Database.BeginTransaction(IsolationLevel.RepeatableRead))
             {
                 var reservation = await db.ReceivingAddressReservations
-                    .Include(r => r.ReceivingAddress)
-                    .FirstAsync(r => r.Id == id);
+                    .Include(r => r.Address)
+                    .SingleAsync(r => r.Id == id);
 
                 if (reservation == null)
```
ultimaweapon: `reservation` will never be `null`.

src/Ztm.WebApi/AddressPools/EntityReceivingAddressStorage.cs
```diff
@@ -109,22 +110,22 @@ public async Task ReleaseAsync(Guid id, CancellationToken cancellationToken)
...
 
         public async Task<ReceivingAddressReservation> TryLockAsync(Guid id, CancellationToken cancellationToken)
         {
             using (var db = this.databaseFactory.CreateDbContext())
-            using (var tx = db.Database.BeginTransaction())
+            using (var tx = db.Database.BeginTransaction(IsolationLevel.RepeatableRead))
             {
-                var recv = await db.ReceivingAddresses.FirstAsync(a => a.Id == id, cancellationToken);
+                var recv = await db.ReceivingAddresses.SingleAsync(a => a.Id == id, cancellationToken);
                 if (recv == null)
```
ultimaweapon: Same here.

src/Ztm.WebApi/AddressPools/EntityReceivingAddressStorage.cs
```diff
@@ -109,22 +110,22 @@ public async Task ReleaseAsync(Guid id, CancellationToken cancellationToken)
...
                 {
-                    return null;
+                    throw new ArgumentNullException("Receiving address id is not found.");
```
ultimaweapon: I think `ArgumentNullException` is not the right exception here.

src/Ztm.WebApi/AddressPools/EntityReceivingAddressStorage.cs
```diff
@@ -152,19 +153,19 @@ public async Task<ReceivingAddressReservation> TryLockAsync(Guid id, Cancellatio
...
-                BitcoinAddress.Create(receivingAddress.Address, this.network),
-                receivingAddress.IsLocked,
+                entity.Id,
+                BitcoinAddress.Create(entity.Address, this.network),
+                entity.IsLocked,
                 new List<ReceivingAddressReservation>()
             );
 
-            if (receivingAddress.ReceivingAddressReservations != null)
+            if (entity.Reservations.Any())
```
ultimaweapon: We don't need this line.

src/Ztm.WebApi.Tests/AddressPools/EntityReceivingAddressStorageTests.cs
```diff
@@ -0,0 +1,261 @@
...
+            // Act.
+            var result = await this.subject.ListAsync(AddressFilter.Available, CancellationToken.None);
+
+            // Assert.
+            Assert.Single(result);
+            Assert.Equal(receivingAddresses.Last().Id, result.First().Id);
+        }
+
+        [Fact]
+        public async Task ListReceivingAddressAsync_WithNeverUsedFlag_ShouldAllFreshAddresses()
```
ultimaweapon: Same as previous issue.

src/Ztm.WebApi.Tests/AddressPools/EntityReceivingAddressStorageTests.cs
```diff
@@ -0,0 +1,261 @@
...
+            // Act & Assert.
+            _ = Assert.ThrowsAsync<InvalidOperationException>(
+                () => this.subject.ReleaseAsync(reservation.Id, CancellationToken.None)
+            );
+        }
+
+        [Fact]
+        public void ReleaseAsync_InExist_ShouldThrow()
+        {
+            _ = Assert.ThrowsAsync<KeyNotFoundException>(
```
ultimaweapon: This not correct. We should `await` instead.

src/Ztm.WebApi.Tests/AddressPools/EntityReceivingAddressStorageTests.cs
```diff
@@ -0,0 +1,261 @@
...
+        [Fact]
+        public async Task ReleaseAsync_Released_ShouldThrow()
+        {
+            // Arrange.
+            var address = await this.subject.AddAsync(TestAddress.Regtest1, CancellationToken.None);
+            var reservation = await this.subject.TryLockAsync(address.Id, CancellationToken.None);
+            await this.subject.ReleaseAsync(reservation.Id, CancellationToken.None);
+
+            // Act & Assert.
+            _ = Assert.ThrowsAsync<InvalidOperationException>(
```
ultimaweapon: Same here.

src/Ztm.WebApi.Tests/AddressPools/EntityReceivingAddressStorageTests.cs
```diff
@@ -0,0 +1,261 @@
...
+using NBitcoin;
+using Xunit;
+using Ztm.Data.Entity.Testing;
+using Ztm.Testing;
+using Ztm.WebApi.AddressPools;
+using Ztm.Zcoin.NBitcoin;
+
+namespace Ztm.WebApi.Tests.AddressPools
+{
+    public sealed class EntityReceivingAddressStorageTests : IDisposable
```
ultimaweapon: We should add cases to check when `TryLockAsync` on non-existence.

src/Ztm.WebApi.Tests/AddressPools/ReceivingAddressPoolTests.cs
```diff
@@ -0,0 +1,149 @@
...
+
+            // Act.
+            var recv = await this.subject.TryLockAddressAsync(CancellationToken.None);
+
+            // Assert.
+            Assert.NotNull(recv);
+        }
+
+        [Fact]
+        public async Task TryLockAddressAsync_AndConsumedAllAvailable_ShouldReturnNullWhenTryToLockNext()
```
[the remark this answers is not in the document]
thebevrishot: We should verify available addresses in pool is reduced when lock success.

src/Ztm.WebApi.Tests/AddressPools/LessUsageFirstChoserTests.cs
```diff
@@ -0,0 +1,90 @@
...
+                    ));
+            }
+
+            // Act.
+            var chosen = this.subject.Choose(availables);
+
+            // Assert.
+            var usage = chosen.Reservations.Count;
+
+            Assert.Equal(expected, usage);
```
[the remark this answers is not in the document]
thebevrishot: The purpose of this testcase is asserting usage amount. We could not expect which address is chose.

src/Ztm.WebApi.Tests/AddressPools/TestReceivingAddressStorage.cs
```diff
@@ -0,0 +1,102 @@
...
+using System.Collections.ObjectModel;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using NBitcoin;
+using Ztm.WebApi.AddressPools;
+
+namespace Ztm.WebApi.Tests.AddressPools
+{
+    public class TestReceivingAddressStorage : IReceivingAddressStorage
```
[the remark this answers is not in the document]
thebevrishot: This class need to be public because it should be able access by `NSubstitute`

src/Ztm.WebApi.Tests/AddressPools/EntityReceivingAddressStorageTests.cs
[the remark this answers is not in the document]
thebevrishot: I believe it is `EF` layer. We should not verify it.

src/Ztm.WebApi.Tests/AddressPools/LessUsageFirstChoserTests.cs
```diff
@@ -0,0 +1,90 @@
...
+            Assert.Throws<ArgumentException>(
+                "addresses",
+                () => this.subject.Choose(new Collection<ReceivingAddress>()));
+        }
+
+        [Theory]
+        [InlineData(0, 0, 0, 1, 1)]
+        [InlineData(1, 2, 2, 1, 1)]
+        [InlineData(999, 999, 1000, 999, 1000)]
+        public void Choose_FromList(int expected, params int[] vs)
```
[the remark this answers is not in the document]
thebevrishot: Good idea

src/Ztm.WebApi.Tests/AddressPools/LessUsageFirstChoserTests.cs
```diff
@@ -0,0 +1,90 @@
...
+                (
+                    Guid.NewGuid(),
+                    TestAddress.Regtest1,
+                    false,
+                    emptyReservations
+                );
+
+                var reservations = new List<ReceivingAddressReservation>();
+
+                foreach (var _ in Enumerable.Range(0, v))
```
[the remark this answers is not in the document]
thebevrishot: `for` can not be used.

src/Ztm.WebApi.Tests/AddressPools/TestReceivingAddressStorage.cs
```diff
@@ -0,0 +1,102 @@
...
+                    ? new List<ReceivingAddressReservation>()
+                    : recv.Reservations;
+
+                var last = reservations.Last();
+                reservations.Remove(last);
+                reservations.Add(new ReceivingAddressReservation(last.Id, last.Address, last.ReservedDate, DateTime.UtcNow));
+
+                var updated = new ReceivingAddress(recv.Id, recv.Address, false, reservations);
+
+                this.receivingAddresses.AddOrReplace(id, updated);
```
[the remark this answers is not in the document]
thebevrishot: Just `Remove` and `Add` is ok.

src/Ztm.WebApi.Tests/AddressPools/TestReceivingAddressStorage.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            if (this.receivingAddresses.TryGetValue(id, out var recv))
+            {
+                if (recv.IsLocked)
+                {
+                    throw new InvalidOperationException();
+                }
+
+                var lockedAt = DateTime.UtcNow;
+                var reservation = new ReceivingAddressReservation(Guid.NewGuid(), recv, lockedAt, DateTime.MinValue);
```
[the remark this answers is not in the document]
thebevrishot: Nice catch

src/Ztm.WebApi.Tests/AddressPools/EntityReceivingAddressStorageTests.cs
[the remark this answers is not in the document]
thebevrishot: same as above

src/Ztm.WebApi.Tests/AddressPools/EntityReceivingAddressStorageTests.cs
```diff
@@ -0,0 +1,261 @@
...
+                await this.subject.AddAsync(TestAddress.Regtest1, CancellationToken.None),
+                await this.subject.AddAsync(TestAddress.Regtest2, CancellationToken.None)
+            };
+
+            // Act.
+            var result = await this.subject.ListAsync(AddressFilter.None, CancellationToken.None);
+
+            // Assert.
+            Assert.Contains(receivingAddresses.First(), result);
+            Assert.Contains(receivingAddresses.Last(), result);
```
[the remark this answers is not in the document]
thebevrishot: Use Subset of `Xunit` instead.

src/Ztm.WebApi.Tests/AddressPools/ReceivingAddressPoolTests.cs
```diff
@@ -87,17 +88,21 @@ public async Task TryLockAddressAsync_AndAllAddressesAreUnavailable_ShouldReturn
...
 
             // Act.
-            var recv = await this.subject.TryLockAddressAsync(CancellationToken.None);
+            var recv = await this.subject.TryLockAddressAsync(cancellationToken);
 
             // Assert.
             Assert.NotNull(recv);
+            _ = this.storage.Received(1).ListAsync(Arg.Any<AddressFilter>(), Arg.Is<CancellationToken>(c => c == cancellationToken));
+            _ = this.choser.Received(1).Choose(Arg.Is<IEnumerable<ReceivingAddress>>(rs => rs.Any()));
+            _ = this.storage.Received(1).TryLockAsync(Arg.Any<Guid>(), Arg.Is<CancellationToken>(c => c == cancellationToken));
```
ultimaweapon: We should assert if it is locked the correct one too.

src/Ztm.WebApi.Tests/AddressPools/ReceivingAddressPoolTests.cs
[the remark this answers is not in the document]
ultimaweapon: Sorry I mean it is duplicated with `TryLockAddressAsync_AndAllAddressesAreUnavailable_ShouldReturnNull`.

src/Ztm.WebApi.Tests/AddressPools/LessUsageFirstChoserTests.cs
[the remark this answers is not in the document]
ultimaweapon: The problems with that way is it will conflict with the purpose of this class, which is selecting the most less usages address. So it should assert the expected address instead.

src/Ztm.WebApi.Tests/AddressPools/LessUsageFirstChoserTests.cs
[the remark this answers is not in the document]
ultimaweapon: This should works:

```csharp
for (int i = 0; i < u; i++)
{
}
```

src/Ztm.WebApi.Tests/AddressPools/EntityReceivingAddressStorageTests.cs
```diff
@@ -84,7 +92,7 @@ public async Task GetAsync_ExistReceivingAddress_ShouldSuccess()
         }
 
         [Fact]
-        public async Task ListReceivingAddressAsync_WithEmptyRecerivingAddress_ShouldGetEmpty()
+        public async Task ListAsync_WithEmptyRecerivingAddress_ShouldGetEmpty_WithEmptyRecerivingAddress_ShouldGetEmpty()
```
ultimaweapon: Seems like there is accidentally paste here.

src/Ztm.WebApi.Tests/AddressPools/EntityReceivingAddressStorageTests.cs
```diff
@@ -107,12 +115,11 @@ await this.subject.AddAsync(TestAddress.Regtest2, CancellationToken.None)
             var result = await this.subject.ListAsync(AddressFilter.None, CancellationToken.None);
 
             // Assert.
-            Assert.Contains(receivingAddresses.First(), result);
-            Assert.Contains(receivingAddresses.Last(), result);
+            Assert.Subset(result.ToHashSet(), receivingAddresses.ToHashSet());
```
ultimaweapon: Seems like `ToHashSet` is introduced by `NBitcoin`. We should not depend on any utilities from NBitcoin that does not related to Bitcoin.

src/Ztm.WebApi.Tests/AddressPools/LessUsageFirstChoserTests.cs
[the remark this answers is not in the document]
thebevrishot: We should not expect address for example if we have usage of each address 0, 0, 1, 1 we could expect only address that have usage amount equal to 0 would be selected but could not know which address is selected.

src/Ztm.WebApi.Tests/AddressPools/LessUsageFirstChoserTests.cs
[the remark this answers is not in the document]
thebevrishot: Oh, I forget this way LoL. using too much magic of CSharp.

src/Ztm.WebApi.Tests/AddressPools/EntityReceivingAddressStorageTests.cs
thebevrishot: After determine `ToHashSet` is in linq

src/Ztm.WebApi/AddressPools/EntityReceivingAddressStorage.cs
```diff
@@ -65,7 +73,12 @@ public async Task<ReceivingAddress> GetAsync(Guid id, CancellationToken cancella
                     .Include(e => e.Reservations)
                     .SingleOrDefaultAsync(r => r.Id == id);
 
-                return recv == null ? null : ToDomain(recv);
+                if (recv == null)
+                {
+                    throw new KeyNotFoundException("The address id is not found.");
+                }
+
+                return ToDomain(recv);
```
ultimaweapon: The previous way should be better for this case due to getting with invalid ID should be common cases.

src/Ztm.WebApi.Tests/AddressPools/EntityReceivingAddressStorageTests.cs
```diff
@@ -78,6 +78,24 @@ public async Task AddAsync_WithNullAddress_ShouldThrow()
...
+            var address = TestAddress.Regtest1;
+            await this.subject.AddAsync(TestAddress.Regtest1, CancellationToken.None);
+
+            await Assert.ThrowsAsync<ArgumentException>(
+                "address",
+                () => this.subject.AddAsync(TestAddress.Regtest1, CancellationToken.None));
+        }
+
+        [Fact]
+        public async Task GetTask_NonExist_ShouldThrow()
```
ultimaweapon: Case name is not correct.

src/Ztm.WebApi/AddressPools/RpcAddressGenerator.cs
```diff
@@ -0,0 +1,28 @@
...
+using NBitcoin;
+using Ztm.Zcoin.Rpc;
+
+namespace Ztm.WebApi.AddressPools
+{
+    public sealed class RpcAddressGenerator : IAddressGenerator
+    {
+        readonly IZcoinRpcClient client;
+
+        public RpcAddressGenerator(IZcoinRpcClient client)
```
ultimaweapon: We should accept factory instead. The reason is `IZcoinRpcClient` should not be a long running connection. It lifetime should be short and underlying connection should be handled by factory.

src/Ztm.WebApi/Startup.cs
```diff
@@ -89,5 +93,14 @@ IZcoinRpcClientFactory CreateZcoinRpcClientFactory(IServiceProvider provider)
                 RPCCredentialString.Parse($"{config.Rpc.UserName}:{config.Rpc.Password}")
             );
         }
+
+        void UseAddressPools(IServiceCollection services)
+        {
+            services.AddTransient<IAddressChoser, LessUsageFirstChoser>();
```
ultimaweapon: Better to register this one as singleton.

src/Ztm.WebApi/Startup.cs
```diff
@@ -89,5 +93,14 @@ IZcoinRpcClientFactory CreateZcoinRpcClientFactory(IServiceProvider provider)
...
+            services.AddTransient<IAddressGenerator, RpcAddressGenerator>();
```
ultimaweapon: Same here.

src/Ztm.WebApi/Startup.cs
```diff
@@ -89,5 +93,14 @@ IZcoinRpcClientFactory CreateZcoinRpcClientFactory(IServiceProvider provider)
...
+            services.AddTransient<IReceivingAddressStorage, EntityReceivingAddressStorage>();
```
ultimaweapon: Same here.

PR #134, opened 2019-12-12
Implement scheduler for Timer

### LINKED ISSUE #27: Make class Ztm.Threading.Timer to be able to test racecondition
because of PR #26 doesn't prove it can prevent `TimerElapsedAsync` to be called after `StopAsync` activate.

ultimaweapon:
Resolve #27.

### DIFF
--- a/src/Ztm.Threading/Timer.cs
+++ b/src/Ztm.Threading/Timer.cs
@@ -3,49 +3,48 @@ using System.Diagnostics;
 using System.Threading;
 using System.Threading.Tasks;
 using Ztm.ObjectModel;
+using Ztm.Threading.TimerSchedulers;
 
 namespace Ztm.Threading
 {
     public sealed class Timer
     {
-        public static readonly TimeSpan MaxDuration = TimeSpan.FromMilliseconds(4294967294);
-        public static readonly TimeSpan MinDuration = TimeSpan.Zero;
-
         readonly Stopwatch stopwatch;
         volatile TimerStatus status;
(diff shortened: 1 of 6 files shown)


src/Ztm.Threading/TimerSchedulers/ThreadPoolScheduler.cs
```diff
@@ -0,0 +1,64 @@
+using System;
+using System.Threading;
+
+namespace Ztm.Threading.TimerSchedulers
+{
+    public sealed class ThreadPoolScheduler : ITimerScheduler
+    {
+        public static readonly TimeSpan MaxDuration = TimeSpan.FromMilliseconds(4294967294);
```
thebevrishot: Why we use this value? I think we could use `TimeSpan.MaxValue` instead.

src/Ztm.Threading/Timer.cs
```diff
@@ -3,79 +3,65 @@
...
-        volatile System.Threading.Timer timer;
+        volatile object schedule;
         volatile int elapsed;
 
-        public Timer()
+        public Timer(ITimerScheduler scheduler = null)
         {
             this.stopwatch = new Stopwatch();
+
+            Scheduler = scheduler ?? DefaultScheduler;
```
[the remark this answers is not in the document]
ultimaweapon: We don't need to use `this` on property or method due to it does not have the same problems as field.

src/Ztm.Threading/TimerSchedulers/ThreadPoolScheduler.cs
ultimaweapon: Added document.

Issue #133, opened 2019-12-12
Downgrade .NET Core from 2.2 to 2.1


ultimaweapon:
Due to 2.2 is going to end of life at 2019-12-23. The reason we cannot do #115 right now is because Arch Linux does not support yet.

- [ ] Implementation [1.5]
- [ ] Review [1]

PR #139, opened 2019-12-16
Downgrade dotnetcore to 2.1

thebevrishot:
closes https://github.com/zcoinofficial/ztm/issues/133


src/Ztm.WebApi/Controllers/BackgroundServiceErrorController.cs
```diff
@@ -40,9 +40,9 @@ public ActionResult BackgroundServiceError()
                     Service = e.Service.FullName,
                     Error = e.Exception.Message,
                     Detail = e.Exception.StackTrace
-                }).ToList();
+                }).AsEnumerable();
```
ultimaweapon: `AsEnumerable` does not do anything. That mean it will not evaluate now.

src/Ztm.WebApi/Controllers/BackgroundServiceErrorProblemDetails.cs
```diff
@@ -0,0 +1,11 @@
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Mvc;
+using Ztm.WebApi.Models;
+
+namespace Ztm.WebApi.Controllers
+{
+    public sealed class BackgroundServiceErrorProblemDetails : ProblemDetails
```
ultimaweapon: This one should go inside `Models` instead.

src/Ztm.Data.Entity.Postgres.Tests/Mapping/UInt256TypeMappingTests.cs
```diff
@@ -0,0 +1,44 @@
+using Xunit;
+using Ztm.Data.Entity.Postgres.Mapping;
+
+namespace Ztm.Data.Entity.Postgres.Tests
```
ultimaweapon: Namespace does not matched with path name.

Issue #140, opened 2019-12-16
Increase timeout for Ztm.Threading.Tests.TimerSchedulers.ThreadPoolSchedulerTests.Schedule_WithNonNullPeriod_ShouldNotOneShot


ultimaweapon:
Sometime it fail on GitHub Actions due to timeout is too short.

- [ ] Implementation [0.5]
- [ ] Review [0.5]

PR #149, opened 2019-12-20
Increase timeout for Ztm.Threading.Tests.TimerSchedulers.ThreadPoolSchedulerTests.Schedule_WithNonNullPeriod_ShouldNotOneShot 

thebevrishot:
closes https://github.com/zcoinofficial/ztm/issues/140

### DIFF
--- a/src/Ztm.Threading.Tests/TimerSchedulers/ThreadPoolSchedulerTests.cs
+++ b/src/Ztm.Threading.Tests/TimerSchedulers/ThreadPoolSchedulerTests.cs
@@ -97,8 +97,8 @@ namespace Ztm.Threading.Tests.TimerSchedulers
             }
         }
 
-        [Fact(Skip = "This test is not reliable due to it need to assume that timer has been elapsed")]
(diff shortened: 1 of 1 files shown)


src/Ztm.Threading.Tests/TimerSchedulers/ThreadPoolSchedulerTests.cs
```diff
@@ -97,26 +97,34 @@ public void Schedule_WithNullPeriod_ShouldOneShot()
...
+            using (var cde = new CountdownEvent(2))
             {
                 // Arrange.
                 var context = new object();
 
                 this.handler.Setup(f => f(It.IsAny<object>()))
-                            .Callback<object>(c => elapsed.Set());
+                            .Callback<object>(c =>
+                            {
+                                lock(cde)
```
ultimaweapon: We should not do this due to `CountdownEvent` is already thread-safe. We should invoke `Signal` directly and ignore `InvalidOperationException` instead.

Issue #92, opened 2019-10-28
Automatic publish Docker image when tagged


ultimaweapon:

ultimaweapon:
Blocked by #91.

thebevrishot:
We need to wait until release to push this anyway we can use example script from here https://github.com/thebevrishot/test-ztm/blob/master/.github/workflows/cd.yml

PR #163, opened 2019-12-30
Add cd configuration on Github Actions

thebevrishot:
closes https://github.com/zcoinofficial/ztm/issues/92

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -53,7 +53,7 @@ jobs:
       uses: azure/docker-login@v1
       with:
         login-server: docker.pkg.github.com
-        username: $GITHUB_ACTOR
+        username: ${{ github.repository }}
         password: ${{ secrets.GITHUB_TOKEN }}
     - name: Build Image
       run: docker build -t docker.pkg.github.com/zcoinofficial/ztm/ztm:${GITHUB_REF##*/} -f $GITHUB_WORKSPACE/src/Ztm.WebApi/Dockerfile release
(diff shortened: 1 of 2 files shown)


thebevrishot:
@ultimaweapon After investigate I think we need 1 - 2 days to create the action my idea is we could let it be duplicated for now and fix it later after first release.

.github/workflows/cd.yml
```diff
@@ -0,0 +1,61 @@
...
+      uses: actions/upload-artifact@v1
+      with:
+        name: migration
+        path: migration
+  docker:
+    name: Build Docker Image
+    needs: build
+    runs-on: ubuntu-18.04
+    steps:
+    - name: Checkout
```
ultimaweapon: Seems like we don't need this.

.github/workflows/cd.yml
```diff
@@ -0,0 +1,61 @@
...
+      with:
+        name: release
+        path: release
+    - name: Fix File Permission
+      run: chmod +x release/Ztm.WebApi
+    - name: Login Docker
+      uses: azure/docker-login@v1
+      with:
+        login-server: docker.pkg.github.com
+        username: $GITHUB_ACTOR
```
ultimaweapon: I think better to use `secrets` so it match with `password`.

.github/workflows/cd.yml
thebevrishot: I believe we need this because we have to access Dockerfile

.github/workflows/cd.yml
thebevrishot: Not sure what field from `secrets` you mean. I updated to `${{ github.repository }}` instead.

.github/workflows/cd.yml
ultimaweapon: I think we should create artifact for that Dockerfile instead.

Issue #172, opened 2020-01-03
Change uint256 endian in PostgreSQL


ultimaweapon:
So it display will be matched with Zcoin.

PR #174, opened 2020-01-06
Change uint256 format to big endian

thebevrishot:
closes https://github.com/zcoinofficial/ztm/issues/172


thebevrishot:
After personal discussion, We can merge it without data migration script if we can finish this before first release.

.github/workflows/ci.yml
```diff
@@ -23,6 +25,13 @@ jobs:
     - name: Build E2E Runner
       shell: cmd
       run: dotnet.exe build src\Ztm.EndToEndTests
+    - name: Install main database
```
ultimaweapon: I think this name quite hard to understand due to there is possibility that PostgreSQL will host additional database. So change this to `Install PostgreSQL` looks better.

.github/workflows/ci.yml
```diff
@@ -23,6 +25,13 @@ jobs:
...
+      shell: cmd
+      run: choco.exe install postgresql --version=10.6.0.20190131 --params '/Password:postgres' --params-global
```
ultimaweapon: We should disable installation progress due to it output too much.

.github/workflows/ci.yml
```diff
@@ -23,6 +25,13 @@ jobs:
...
+    - name: Migrate main database
```
ultimaweapon: Same here. My idea is `Create Main Database` looks better due to we did not migrate anything.

.github/workflows/ci.yml
```diff
@@ -23,6 +25,13 @@ jobs:
...
+      shell: cmd
+      working-directory: ./src/Ztm.Data.Entity.Postgres
```
ultimaweapon: Path separator in Windows should be `\`. Also I think we can omit `.` at the beginning.

src/Ztm.Data.Entity.Postgres.Tests/TheoryWhenEnvIsSetAttribute.cs
```diff
@@ -0,0 +1,16 @@
+using System;
+using Xunit;
+
+namespace Ztm.Data.Entity.Postgres.Tests
+{
+    public class TheoryWhenEnvIsSetAttribute : TheoryAttribute
```
ultimaweapon: IMO class name look weird. My idea is rename it to `ConditionalTheoryAttribute` with `RequiredEnv` property to specify the property name.

src/Ztm.Data.Entity.Postgres.Tests/MainDatabaseTests.cs
```diff
@@ -0,0 +1,98 @@
...
+                }
+
+                // Get uint256 back.
+                var block = await this.subject.Blocks.FirstOrDefaultAsync();
+                Assert.NotNull(block);
+                Assert.Equal(parsed, block.Hash);
+            }
+            finally
+            {
+                this.subject.Database.ExecuteSqlCommand("TRUNCATE TABLE \"Blocks\" CASCADE");
```
ultimaweapon: We should create a database fixture and move this to it instead.

.github/workflows/ci.yml
thebevrishot: I'm not sure this field comply the Windows directory format and can't see the spec. However my Visual Studio Code show error message when I use Windows's format. This is the error message `String does not match the pattern of "^(\.[^\\?%*:|"<>\n]+)$".`

src/Ztm.Data.Entity.Postgres.Tests/TheoryWhenEnvIsSetAttribute.cs
thebevrishot: Good idea

src/Ztm.Data.Entity.Postgres.Tests/MainDatabaseTests.cs
thebevrishot: Done

.github/workflows/ci.yml
ultimaweapon: Have you tried this? `src\Ztm.Data.Entity.Postgres`

thebevrishot: It show the same error

src/Ztm.Data.Entity.Postgres.Tests/MainDatabaseFixture.cs
```diff
@@ -0,0 +1,74 @@
...
+        public void CleanUp()
+        {
+            // Truncate all tables in schema public except __EFMigrationsHistory
+            var tables = this.ExecuteSql("SELECT table_name FROM information_schema.tables WHERE table_schema = 'public' AND table_name <> '__EFMigrationsHistory';")
+                             .Select(r => r[0] as string);
+
+            using (var db = this.CreateMainDatabase())
+            {
+                var query = "TRUNCATE TABLE " + string.Join(", ", tables.Select(t => '"' + t + '"'));
+                db.Database.ExecuteSqlCommand(query); // lgtm [cs/second-order-sql-injection]
```
ultimaweapon: I think we should use `DbConnection` directly for executing raw SQL.

src/Ztm.Data.Entity.Postgres.Tests/MainDatabaseFixture.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        public void Dispose()
+        {
+            this.connection.Dispose();
+        }
+
+        public IEnumerable<IDataRecord> ExecuteSql(string rawSql)
+        {
+            using (var db = this.CreateMainDatabase())
+            using (var command = db.Database.GetDbConnection().CreateCommand())
```
ultimaweapon: Same here. We can create `DbCommand` from `DbConnection` directly.

src/Ztm.Data.Entity.Postgres.Tests/MainDatabaseFixture.cs
```diff
@@ -0,0 +1,74 @@
...
+                db.Database.OpenConnection();
+                using (var result = command.ExecuteReader())
+                {
+                    while (result.Read())
+                    {
+                        yield return result;
+                    }
+                }
+
+                db.Database.CloseConnection();
```
ultimaweapon: Seems like this one is going to close underlying connection, which mean is will cause subsequence operation to fail due to we have only one connection.

src/Ztm.Data.Entity.Postgres.Tests/MainDatabaseFactoryTests.cs
```diff
@@ -0,0 +1,45 @@
...
+            act = () => new MainDatabaseFactory(null);
+            act.Should().Throw<ArgumentNullException>()
+               .And.ParamName.Should().Be("config");
+        }
+
+        [Fact]
+        public void CreateDbContext_ShouldSuccess()
+        {
+            // Act.
+            var context = this.subject.CreateDbContext();
```
ultimaweapon: We forgot to dispose this.

src/Ztm.Data.Entity.Postgres.Tests/MainDatabaseFixture.cs
thebevrishot: That's right.

src/Ztm.Data.Entity.Postgres.Tests/MainDatabaseFixture.cs
thebevrishot: Good idea

src/Ztm.Data.Entity.Postgres.Tests/MainDatabaseFixture.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        public MainDatabaseFixture()
+        {
+            var connectionString = Environment.GetEnvironmentVariable("ZTM_MAIN_DATABASE");
+            if (connectionString == null)
+            {
+                throw new Exception("No ZTM_MAIN_DATABASE environment variable is set.");
+            }
+
+            this.connection  = new NpgsqlConnection(connectionString);
```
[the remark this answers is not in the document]
thebevrishot: fixed

